PR #2, opened 2018-03-19
[11.0][ADD] account_fx_spot

LoisRForgeFlow:
Moved from https://github.com/OCA/account-financial-tools/pull/619

Account FX Spot
===============

This module extends the functionality of Accounting to support Foreign
Exchange Spot Transactions. A **Foreign Exchange Spot Transaction**, also
known as FX spot, is an agreement between two parties to buy one currency
against selling another currency at an agreed price for settlement on
the spot date. The exchange rate at which the transaction is done is
called the spot exchange rate.

Configuration
=============

To configure this module, you need to:

#. Go to *Accounting > Configuration > Settings*.
#. Enable the *Multi-Currencies* option.

Usage
=====

With this module, you will be able to:

* Create FX Spot Transactions.

#. Go to *Accounting > Foreign Exchange > FX Spot Transaction*.
#. Select a Partner and fill the outgoing and incoming amounts and currencies.

* Register Payments for FX Spot Transactions.

#. Select Several transactions from the tree view.
#. Click on *Action > Register Payment*.

or

#. Go to the form view for an open transaction.
#. Click on *Register Payment*.


JordiBForgeFlow (approved):

JordiBForgeFlow:
@Yenthe666 you wanted this one, right? Your review is welcome!

account_fx_spot/models/account_fx_spot.py
```diff
@@ -0,0 +1,354 @@
...
+    @api.onchange("amount_out", "amount_in")
+    def _onchange_amounts(self):
+        self.rate = \
+            self.amount_in / self.amount_out if self.amount_out else 0.0
+
+    @api.onchange("rate")
+    def _onchange_rate(self):
+        self.amount_in = self.amount_out * self.rate
+
+    def _prepare_account_move(self):
```
Yenthe666: decorator?

account_fx_spot/models/account_fx_spot.py
```diff
@@ -0,0 +1,354 @@
...
+
+    @api.multi
+    def action_re_open(self):
+        if self.filtered(lambda r: r.state != "done"):
+            raise UserError(_(
+                "FX Spot Transaction must be done in order to set it to "
+                "register payment."))
+        return self.write({"state": "open"})
+
+    def action_cancel(self):
```
Yenthe666: decorator?

Yenthe666:
Hey @jbeficent,

Just a fast & highlevel review but it does look good.
P.S: do you guys never add the `.po` files?

JordiBForgeFlow:
@Yenthe666 .po files are never added in OCA modules. Transifex is able to do the translations without them.

account_fx_spot/models/account_fx_spot.py
LoisRForgeFlow: api.multi is the default, right? but yeah maybe for consistency with the other methods in the file I should also explicitly set it for theese methods.

JordiBForgeFlow:
Please squash your commits.

LoisRForgeFlow:
@jbeficent I was waiting for @Yenthe666 to update his review. Diff could make that very quick for him!

account_fx_spot/models/account_fx_spot.py
Yenthe666: @lreficent yes it is the default. Some developers do not add it because it is 'the default', but what if at some point Odoo changes the API and this behaviour? Then you will have random functions breaking because of this.
I know this is a bug if and it feels like neatpicking but after all it only takes a few seconds to add an `@api.multi` right? 😉 
Thanks for adding!

Yenthe666 (approved):

Yenthe666:
Thanks a lot, it does make it a lot easier. It looks good to me now, thanks for adding the decorators!

JordiBForgeFlow:
@lreficent please squash and I will.merge.

account_fx_spot/models/account_fx_spot.py
LoisRForgeFlow: Yes, that's a fair point. Of course no problem to change it! Thanks

LoisRForgeFlow:
@jbeficent Done!

JordiBForgeFlow:
Did you check the compatibility with currency_rate_inverted?

JordiBForgeFlow (approved):

feketemihai:
@lreficent Can you please adjust a little bit the view, seems a little bit strange and quite hard to use, by default you expect to input first the input amount and just after output amount,
 And i have an error when push save button:

The operation cannot be completed, probably due to the following:
- deletion: you may be trying to delete a record while other records still reference it
- creation/update: a mandatory field is not correctly set

[object with reference: out_currency_id - out.currency.id]

fxspot

JordiBForgeFlow:
@feketemihai you have to provide both incoming and outgoing currency. Can you provide us the steps to reproduce the error?

feketemihai:
@jbeficent i made the screenshot just for rearrange the fields. The error was with both currencies set, tested with currencies different than company currency, one equal to company currency, and the error was the same.

LoisRForgeFlow:
@feketemihai I got the error, it is weird, I'm checking. 

About the screen, on my local the Incoming group was just at the right of the spot exchange group.

feketemihai:
@lreficent The screenshot is from runbot.

account_fx_spot/views/account_fx_view.xml
```diff
@@ -80,6 +79,17 @@
                         </group>
                         <group name="rate" string="Spot Exchange Rate">
                             <field name="rate" nolabel="1"/>
+                            <div>
+                                <field name="in_currency_id" readonly="1"/>
+                                /
+                                <field name="out_currency_id" readonly="1"/>
+                            </div>
```
LoisRForgeFlow: The issue are these divs where I use again in_currency_id and out_currency_id.

The desired effect is this:
image

What other options do I have? compute a string field to display it? any better idea?

LoisRForgeFlow:
@feketemihai yes, I saw it on runbot too, let's first discuss the error, then the UI. See my comment above.

account_fx_spot/views/account_fx_view.xml
feketemihai: @lreficent Since rates are just for information i suggest to change those one to text and compute rate + in_currency + out_currency (and inverse), otherwise you have to add another fields for getting the currency names, so in total 4 fields, and to adjust tests after not to check float value of rate.

LoisRForgeFlow: just typed "outgoing" and "incoming", I think that works for now. I also switched the position of the incoming and outgoing groups as you suggested.

feketemihai:
@lreficent Can you check, if i use different currencies that the payment journal, the outgoing payment is fine but for incoming it raises error that: Incoming Currencies does not match. It should be possible as well just like the outgoing...

LoisRForgeFlow:
@feketemihai You are right, that is not supported at the moment but it would involve some tricky development into the account.payment logic. Plus, I don't think it would be the most common case in a FX spot transaction, specially for the incoming currency which is the one you are buying so you are supposed to have a specific bank/journal for it.

I added the issue in the README.

feketemihai:
@lreficent Thanks for the issue and for this great module.

feketemihai (approved):

feketemihai:
@pedrobaeza This one too, can be merged...Thanks.

pedrobaeza:
@lreficent please squash before last 3 commits before merging

LoisRForgeFlow:
@pedrobaeza Done!

PR #3, opened 2018-03-19
[11.0][add] currency_rate_inverted

JordiBForgeFlow:
Currency Exchange Rate Inverted
===============================

This module applies the inverse method in the conversion between two currencies.

The inverse method is a calculation method that uses the inverse (reciprocal)
exchange rate for the multiplier and divisor when converting amounts from one
currency to another.

The exchange rate of the multiplier and divisor are the reciprocal,
or opposite, of each other.

The inverse method multiplies the foreign amount by the
exchange rate to calculate the company currency (also called domestic) amount.

Configuration
=============

To convert amounts from one currency to another using the inverse method,
you have to go to 'Accounting / Configuration / Miscellaneous / Currencies'.
Then select the foreign currency that you wish to maintain inversion for
and set the flag 'Inverted exchange rate'.

Supersedes https://github.com/OCA/account-financial-tools/pull/627


currency_rate_inverted/README.rst
```diff
@@ -0,0 +1,76 @@
...
+or opposite, of each other.
+
+The inverse method multiplies the foreign amount by the
+exchange rate to calculate the company currency (also called domestic) amount.
+
+Configuration
+=============
+
+To convert amounts from one currency to another using the inverse method,
+you have to go to 'Accounting / Configuration / Miscellaneous / Currencies'.
```
LoisRForgeFlow: `Accounting / Configuration / Accounting / Currencies`

LoisRForgeFlow (approved):
Tested on runbot, works properly :+1:

JordiBForgeFlow:
@jcdrubay @huyly0909 this is ready for your review.

JordiBForgeFlow:
Well, in fact the module was proposed by @huyly0909 in the first place, and I have reviewed and validated, as well as @lreficent. So merging now..

PR #5, opened 2018-03-19
[11.0][mig][currency_rate_update]

JordiBForgeFlow:
Originating from https://github.com/OCA/account-financial-tools/pull/579


JordiBForgeFlow:
The module was first proposed by @jesusVMayor and approved by @feketemihai and myself. So moving on and merging this one.

pedrobaeza:
@jbeficent again you are merging without squashing useless commits like your last ones. Please ping me before merging and I can do the squash, but don't merge directly.

feketemihai:
@jbeficent @pedrobaeza Original PR contained splitted modules for each service that was available in V10, i don't see them merged in this repo, how we should treat that ones? New PR here or it should take place in each localization based on the country of the service?

pedrobaeza:
IMO, if they are local to the country, then in the localization repository. If not, new PR here.

feketemihai:
@pedrobaeza Most of them where from countries central banks, so to each localisation then...Thanks.

PR #6, opened 2018-03-19
[11.0][mig][currency_rate_update_inverted]

JordiBForgeFlow:
Moved from https://github.com/OCA/account-financial-tools/pull/628


JordiBForgeFlow:
@huyly0909 using https://github.com/OCA/currency/pull/5/files#diff-8e3e558fda291236aef336860358c181R177 one can get rid of the extra dependency.

PR #12, opened 2018-03-21
[11.0] [IMP] Update currency_rate_update config view

feketemihai:
Update accounting config view to show auto_currency_up field and label, and make enterprise version checkbox invisible.

### DIFF
--- a/currency_rate_update/views/res_config_settings.xml
+++ b/currency_rate_update/views/res_config_settings.xml
@@ -10,12 +10,15 @@
         <field name="model">res.config.settings</field>
         <field name="inherit_id" ref="account.res_config_settings_view_form"/>
         <field name="arch" type="xml">
-            <field name="module_currency_rate_live" position="attributes">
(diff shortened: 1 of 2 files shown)


currency_rate_update/views/res_config_settings.xml
```diff
@@ -13,9 +13,15 @@
             <field name="module_currency_rate_live" position="attributes">
                 <attribute name="invisible">1</attribute>
             </field>
+            <label for="module_currency_rate_live" position="attributes">
```
pedrobaeza: You shouldn't do in this module. If you don't want that, use `res_config_enterprise_remove` (not sure the exact name) module.

feketemihai: I updated the view and left both options...

pedrobaeza (approved):

JordiBForgeFlow (approved):

PR #23, opened 2018-06-18
[ADD][11.0] [currency_rate_update] - EE, BY services

nedaszilinskas:


pedrobaeza:
You should include this new services as separate modules. Are they specific for a country? If so, I think they should be in the corresponding localization.

PR #28, opened 2018-07-24
[11.0][MIG] currency_rate_date_check

MaxyMoos:
I migrated the module `currency_rate_date_check` to V11.

A few tests were added and a new feature too : if no rates exist in database, a warning is also triggered (unless the considered currency is the company currency of course)


JordiBForgeFlow:
Thanks for proposing the PR!

Can you adjust to the manifest and README formatting features? https://github.com/OCA/maintainer-tools/tree/master/template/module

MaxyMoos:
@jbeficent Sure thing, I pushed a new commit to update README & manifest.
I'll try to work out what makes tests fail in Travis while they pass on a local environment & get back to you.

MaxyMoos:
I've dropped the tests I had written as it would seem that the DB state when running the tests through Travis would make them fail (on a blank DB / vanilla Odoo on my machine, they did run OK).

The failure still seen on Travis seems to come from the fact that the installation of this module adds some rules over currencies that makes other modules' tests fail.

MaxyMoos:
Any news/info about what can or should be done in order to get this PR going forward ?

LoisRForgeFlow:
@MaxyMoos Please, Could you check and fix travis errors? Some tests are failling.

LoisRForgeFlow:
> The failure still seen on Travis seems to come from the fact that the installation of this module adds some rules over currencies that makes other modules' tests fail.

I see, you should try to make the module to integrate properly.

PR #30, opened 2018-09-19
[11.0] [ADD] currency_rate_update_ch_admin

angelmoya:
This module introduces the ability to download the exchange rates associated to currencies from https://admin.ch


angelmoya:
Admin.ch webservices is not working from several weeks.

PR #32, opened 2018-10-02
[12.0][MIG] currency_rate_update

rven:


pedrobaeza:
Please check runbot warning

gdgellatly:
@pedrobaeza What do you suggest, its kind of the way the module works.  Would look funny to change one of the labels in UI.  Maybe "Available Currencies" and "Currencies Available"?

EDIT: By the way, I'm only here because I tried to review in a git-aggregator file, but the ref errorred with no common ancestor.

pedrobaeza:
Yes, the warning is:
```
2018-10-13 09:56:25,724 151 WARNING openerp_test odoo.addons.base.models.ir_model: Two fields (currency_list_readonly, currency_list) of currency.rate.update.service() have the same label: Currencies available
```
So I suppose this is a kind of trick for having the same value 2 times (through a related one), hide one or the other depending of a condition and because of previously there weren't support for sending back values to server if they are `readonly`. But now you can add `force_write` option in the view and the value will be sent back, so no need of 2 files and 1 can be removed.

gdgellatly:
I think its `force_save` but yes I agree.

gdgellatly:
@rven are you coming back to this?

rven:
@gdgellatly I'm not working on it at the moment.

It is not just the two fields you mention, there is also a third field **currency_to_update**

`    # List of currencies available on webservice

    currency_list = fields.Many2many('res.currency',
                                     'res_currency_update_avail_rel',
                                     'service_id',
                                     'currency_id',
                                     string='Currencies available')

    # I can't just put readonly=True in the field above because I need
    # it as r+w for the on_change to work

    currency_list_readonly = fields.Many2many(
        related='currency_list', readonly=True)

    # List of currency to update

    currency_to_update = fields.Many2many('res.currency',
                                          'res_currency_auto_update_rel',
                                          'service_id',
                                          'currency_id',
                                          string='Currencies to update with '
                                          'this service')`

PR #34, opened 2018-10-22
[11.0][add][currency_iso_4217]

JordiBForgeFlow:
Currency ISO 4217
==============
This module adds the following fields to the currencies, as defined by the the
ISO 4127 standard:

* Numeric Code

* Full Name

The module also updates the existing currencies in Odoo,
based on https://github.com/datasets/currency-codes


pedrobaeza:
Is it full name part of the ISO standard?

JordiBForgeFlow:
@pedrobaeza IMHO community-data-files is intended to be used for large data sets, which is not the case. If we have a very specific repo OCA/currency, people will be glad that everything related is contained here.

As for this module adding new currencies, I am not sure that they will be needed. And if so, why not request to add them to upstream directly?

JordiBForgeFlow:
@pedrobaeza Yes, full name is part of the ISO standard.

pedrobaeza:
@jbeficent this is the same large dataset as iso 3166 for countries, and it's not a question of size, but to put all data sets in the same place, that is intended for that. Take into account that this module is not used by any other module in OCA/currency, but for one in OCA/l10n-spain instead, so there's also no correlation of the use of the module with currency related tasks (only that the data it adds is on res.currency model).

pedrobaeza:
Any way, it's not critical and you know that sometimes a module can be hosted in several repositories as they are not totally independent. I won't discuss anymore if you want to host it here. It was just my way of thinking for this specific thing.

JordiBForgeFlow:
@pedrobaeza ok, will move it to be consistent then

JordiBForgeFlow:
Moving the module to https://github.com/OCA/community-data-files/pull/43

PR #36, opened 2018-10-22
[11.0][add] currency_iso_4217

JordiBForgeFlow:
Currency ISO 4217
==============
This module adds the following fields to the currencies, as defined by the the
ISO 4127 standard:

* Numeric Code

* Full Name

The module also updates the existing currencies in Odoo,
based on https://github.com/datasets/currency-codes

### DIFF
--- a/currency_iso_4217/__init__.py
+++ b/currency_iso_4217/__init__.py
@@ -1,0 +1,1 @@
+from . import models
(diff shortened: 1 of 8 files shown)


pedrobaeza:
Maybe you should put this in OCA/community-data-files and include that data for all the currencies.

JordiBForgeFlow:
@pedrobaeza Moved discussion to https://github.com/OCA/currency/pull/34

PR #37, opened 2018-10-25
FIX next_run to prevent update fail

angelmoya:
If one day cron fails, then next_run is not updated, and refresh_currency will not be executed anymore


pedrobaeza (approved):

angelmoya:
Done #724

PR #38, opened 2018-12-17
[IMP] Make the method get_exchange_rate overridable to adapt some special banks

novawish:
Original issue from module [currency_rate_update_vn_VCB](https://github.com/OCA/l10n-vietnam/pull/8):

[VCB service](http://www.vietcombank.com.vn/ExchangeRates/ExrateXML.aspx) already provided the rates as inverted compared with main currency (VND).

So when using this module along with `currency_rate_inverted`, it does not work properly with the inverted rate (inverts the rates that were already inverted).

By providing the overridable mechanism, we'll allow people to handle some specific cases for specific bank services easier.

### DIFF
--- a/currency_rate_update/README.rst
+++ b/currency_rate_update/README.rst
@@ -84,6 +84,8 @@ Contributors
 * Daniel Dico <ddico@oerp.ca> (BOC)
 * Dmytro Katyukha <firemage.dima@gmail.com>
 * Jesús Ventosinos Mayor <jesus@comunitea.com>
+* Duc, Dao Dong <duc.dd@komit-consulting.com> (https://komit-consulting.com)
+* Le Dinh Tien <tien.ld@komit-consulting.com> (https://komit-consulting.com)
 
 Maintainer
 ----------
--- a/currency_rate_update/__manifest__.py
+++ b/currency_rate_update/__manifest__.py
@@ -2,7 +2,7 @@
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
 {
     "name": "Currency Rate Update",
-    "version": "11.0.1.0.0",
+    "version": "11.0.1.0.1",
     "author": "Camptocamp,Odoo Community Association (OCA)",
     "website": "http://camptocamp.com",
     "license": "AGPL-3",
--- a/currency_rate_update/models/currency_rate_update.py
+++ b/currency_rate_update/models/currency_rate_update.py
@@ -171,6 +171,30 @@ class CurrencyRateUpdateService(models.Model):
                             ('name', '=', rate_name)])
                         if not rates:
                             rate = res[curr.name]
+                            if type(rate) is dict:
+                                inverted = rate.get('inverted', None)
+                                direct = rate.get('direct', None)
+                                if inverted is None and direct is None:
+                                    raise UserError(
+                                        _(
+                                            'Invalid rate from %(provider)s '
+                                            'for %(currency)s : %(rate)s'
+                                        ) % {
+                                            'provider': self.service,
+                                            'currency': curr.name,
+                                            'rate': rate,
+                                        }
+                                    )
+                                elif inverted is None:
+                                    inverted = 1/direct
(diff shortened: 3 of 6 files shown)


pedrobaeza (changes requested):
Please include in the method all the code for getting the rate (the search and so on), not only the part you are interested in.

novawish:
Sorry I'm not really sure what you mean.
I split that part to a new method not because i like it, but it would be easier for another module to override it thus calling the super and only change the necessary code that affects the exchange rate.
Can you explain more details about how do we make an override-able method?

pedrobaeza:
I'm saying to extract from line 179 to 190.

novawish:
@pedrobaeza Done, can you please check?

novawish:
BTW, i'm not sure travis's failure is related to this change (the automatic tests failed on module `account_fx_spot`)
image

pedrobaeza:
You have a flake8 error: `currency_rate_update/models/currency_rate_update.py:161:9: F841 local variable 'rate_obj' is assigned to but never used`, but the rest is a change upstream. @jbeficent were you the one that added `account_fx_spot` module?

Syentystech1:
@jbeficent could you please answer @pedrobaeza ? We are a bit blocked, waiting for that PR.  Thanks :)

novawish:
@pedrobaeza I've done the modifications following your suggestion. However after thinking for a long time i think it's really not convenient for other modules to override the function in order to interfere the method generating the rates (for example with current customization i had to override the whole function and could not call super). Can you help me on this?

Thanks very much.

pedrobaeza:
@novawish Indeed it seems a bit weird that you need to override this basic method, but I didn't want to judge.

alexey-pelykh:
In #42 for v12 there has been a refactor that would allow banks to provide both rates or just inverted or just direct, if I got this issue correctly

novawish:
@alexey-pelykh You got it correctly.
However what i wonder is if any chances that the solution would be backported to version 11?
Thanks so much for your comment.

alexey-pelykh:
@novawish well, "as-is" very unlikely, since it would mean that all already-ported currency providers available would have to be refactored as well. For v12 it's easier since nothing is yet ported. Yet it would be possible to grab same approach from #42 and to make less heavy change in the module itself. Idea is that if `rate = res[curr.name]` is scalar value, treat is as before, yet if it's a dictionary with `inverted` key and currency is inverted - don't do the inversion magic. I believe that's the only way to get at least something backported.

novawish:
@alexey-pelykh @pedrobaeza I updated the method to generate the rates, please help me to check.

BTW, travis failed apparently because of module `account_fx_spot`, so i don't know how i can fix it

novawish:
@alexey-pelykh @pedrobaeza Travis and runbot are green now, it would be great if you let some reviews.

alexey-pelykh:
@novawish you may take a look on how this module was reworked for v12, there are extra cases handled of cross-currency inversions.

alexey-pelykh (changes requested):
Please cover your change with tests

novawish:
@alexey-pelykh It's all green now :)

alexey-pelykh (approved):
Code LGTM, yet this would need manual testing "just to be sure" - which I haven't done

novawish:
@alexey-pelykh Well, of course it was manually tested on my local. AFAIK there's only one service that could take advantage from this change is [currency_rate_update_vn_VCB](https://github.com/OCA/l10n-vietnam/pull/8) (is also the reason why i made this PR as mentioned earlier from the beginning).

So to say, if this PR is merged, we will proceed to adapt the module above to use this new implementation.

Thanks very much for all of your help, you did brighten me a lot :)

novawish:
@pedrobaeza May I get your kind review on this? Thank you.

pedrobaeza (approved):

pedrobaeza:
/ocabot merge

I haven't tested myself, but trusting on your good knowledge :smiley:

PR #41, opened 2019-02-13
[12.0][MIG] currency_rate_inverted

alexey-pelykh:


alexey-pelykh:
@jbeficent @aheficent please review, if you don't mind

alexey-pelykh:
@jbeficent @aheficent is there a chance to obtain your reviews? :)

jcdrubay (approved):

currency_rate_inverted/__manifest__.py
```diff
@@ -0,0 +1,22 @@
...
+    'summary': 'Allows to maintain an exchange rate using the inversion '
+               'method',
+    'author': 'Eficent,Odoo Community Association (OCA)',
+    'website': 'https://github.com/OCA/currency',
+    'license': 'AGPL-3',
+    'depends': ['base'],
+    'data': [
+        'views/res_currency_view.xml'
+    ],
+    'test': [],
```
tbaden: if the tag is empty just delete it.

tbaden (changes requested):
thanks for your work. just a minor detail (see below).

 runbot is rebuilding right now. will test this as well when finished.

currency_rate_inverted/__manifest__.py
alexey-pelykh: Fixed

tbaden (approved):
code review: lgtm now

tested in rundeck as well.
inverted currency1 inverted currency2
False False
True False
False True
True True

all combinations calculated correctly

novawish:
> code review: lgtm now
> 
> tested in rundeck as well.
> inverted currency1 inverted currency2
> False False
> True False
> False True
> True True
> 
> all combinations calculated correctly

Did you try these combinations for 2 currencies that are different from the main currency (company's currency)?

Because I just used this module for my project that have: USD, EUR (both are inverted), and VND (main currency).

The rates of USD and EUR compared to VND were correct. However when I do the conversion from EUR to USD apparently the rate is wrongly inverted as the picture below
image

There's no way that EUR is smaller than USD right?
The reason why both EUR and USD are inverted because VND is a very small currency (1 USD ~ 23000 VND).

In general, i think this module does not handle the case above.
I'm not sure that I did something wrong in the configuration, could somebody help me about that?

jcdrubay:
Would this do the trick ?

```python


    @api.model
    def _get_conversion_rate(self, from_currency, to_currency, company, date):
        res = super(ResCurrency, self)._get_conversion_rate(from_currency,
                                                            to_currency)

        if (not from_currency.rate_inverted and not to_currency.rate_inverted):
            # Everything normal
            pass
        elif (from_currency.rate_inverted and to_currency.rate_inverted):
            res = 1 / res
        else:
            currency_rates = (from_currency + to_currency)._get_rates(
                company, date)

            if from_currency.rate_inverted:
                rate_from = 1 / currency_rates.get(from_currency.id)
            else:
                rate_from = currency_rates.get(from_currency.id)

            if to_currency.rate_inverted:
                rate_to = 1 / currency_rates.get(to_currency.id)
            else:
                rate_to = currency_rates.get(to_currency.id)

            res = rate_to / rate_from

        return res

```

alexey-pelykh:
@novawish please check updated version

novawish (approved):
Looks good! Thanks for taking care of my comment.

alexey-pelykh:
@pedrobaeza ?

pedrobaeza:
Blindly merging due to the approvals of other contributors.

novawish:
@alexey-pelykh Since the implementation before did not handle those cases, can we do back-porting for the previous versions?

PR #42, opened 2019-02-13
[12.0][MIG] currency_rate_update

alexey-pelykh:
image
image
image


alexey-pelykh:
@legalsylvain @alexis-via I need your help here with understanding the `max_delta_days` field. For example, "today" is Sunday, and an update is performed. The newest rates available from service are dated as Friday. So in theory, we should save them with date = Friday. What role `max_delta_days` plays in such scenario?

legalsylvain:
Hi @alexey-pelykh. I never used this module and can not help you on that point.

kind regards.

pedrobaeza:
@legalsylvain sorry, I mis-remember that you were one of the guys that touched this and point Alexey to you. @alexis-via for sure modified this though.

alexey-pelykh:
This module is out of WIP stage, please review

@pedrobaeza the refactoring is huge, please take a look as well

kirca:
Hi @alexey-pelykh,

This changes the interface for all existing currency update services. Why is this redesign necessary?

Thanks

alexey-pelykh:
Hi @kirca,

The main reason is: original design exposed a `f(freshness) = rate` that would return newest available data, not older than `freshness` days. New design expects a `f(date) = { actual_date: rate }` implementation, making it possible to do historical queries and allowing service to store `rate` on a valid date, not the query date. As a result, data is not time-consistent.

This allowed to implement "get historical data" action to download exchange rates e.g. for last YTD, if service supports that.

On top of that, the module design didn't follow Odoo-way, if I may say so.

elicoidal:
Too many modules depends on this one: we need to be extremely careful before changing the function signature or data model.
Is there a way to have both functions/structure available at least for v12 and deprecate the old one in v13 with a proper warning in the code?
@alexey-pelykh cc @kirca

alexey-pelykh:
@elicoidal thanks for sharing the concern, I'm aware that some number of modules depend on this one, though it has not yet been ported, thus it's safe to do the transition now as-is, and migrators of the dependant modules (including me) would do the rest. As an alternative, we'd have to issue a new name for this one already and someone would step up to make the legacy module migration, yet taking into account for how long it was not addressed - I don't think that would happen at all.

elicoidal:
@alexey-pelykh fair point

kirca:
The previous design was good as the `CurrencyGetterInterface` was decoupled from Odoo and makes at the very least migrations easier. Also it has a nice dispatching method so there is no need to make explicit checks which is the case when doing it with overriding methods.

I think we should refactor this PR so that we keep the old interface along with the new features. The signature for `get_updated_currency` can change to:

```
get_updated_currency(currency_array, main_currency, max_delta_days=False,
                     date_start=False, date_end=False)
```

Also the return value will now be always a dict:

```
{
    '2019-03-11': {'FOO': 1.5},
    '2019-03-10': {'FOO': 1.45},
}
```
I can do the refactoring if it is needed.
What do you think @alexey-pelykh, @elicoidal?

alexey-pelykh:
@kirca I don't see reason for decoupling Odoo module architecture from Odoo. Why do you suggest this module should be something special vs other modules that follow common pattern and architecture?

kirca:
@alexey-pelykh , I can see a few benefits for having the interface decoupled here:

- easier to see what a new service needs to implement
- easier migrations
- less error prone, i.e. no need to worry about calling `super` if more than one provider service is used and no need to check `if code == "foo"`

alexey-pelykh:
@kirca please see comments below

> easier to see what a new service needs to implement

the sample provided within the module is pretty clear, yet let's add that to USAGE.rst for better understanding.

> easier migrations

less-effort migrations, to be exact.

> less error prone, i.e. no need to worry about calling super if more than one provider service is used and no need to check if code == "foo"

there's a counter-argument: adding another architectural pattern makes it more error-prone already. Regarding the concern mentioned: anyone can forget to call super() in method, and it's very odd to target developers who don't test their code and tests will fail, actually.

kirca:
@alexey-pelykh , additionally testing for new service is easier since it can be done without creating a `res.currency.rate.provider` record.

alexey-pelykh:
@kirca as an counterargument, original approach did not allow custom settings per-provider. For testing there are tests and runbot :)

kirca:
@alexey-pelykh I think we gave enough arguments so lets invite others to judge :)

CC @alexis-via @nbessi @jgrandguillaume @grzegorz-openglobe @yvaucher @guewen @eLBati @jbaudoux

alexey-pelykh:
@kirca Couldn't agree more :)

currency_rate_update/models/res_currency_rate.py
```diff
@@ -0,0 +1,26 @@
+# Copyright 2019 Brainbean Apps (https://brainbeanapps.com)
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html).
+
+from odoo import models, fields, api
+
+
+class ResCurrencyRate(models.Model):
+    _name = 'res.currency.rate'
+    _inherit = ['res.currency.rate', 'mail.thread']
```
pedrobaeza, line 9: I think tracking the rate changes is not adding too much, but overheading everything, so I would remove this inheritance and remove the `track_visibility`.

currency_rate_update/models/res_currency_rate.py
```diff
@@ -0,0 +1,26 @@
...
+        string='Provider',
+        comodel_name='res.currency.rate.provider',
+        ondelete='restrict',
+        track_visibility='onchange',
+    )
+
+    @api.multi
+    def write(self, values):
+        if 'rate' in values and 'provider_id' not in values:
+            values['provider_id'] = False
```
pedrobaeza, line 27: Why do you need this?

currency_rate_update/models/res_currency_rate_provider.py
```diff
@@ -0,0 +1,313 @@
...
+from odoo.exceptions import UserError
+
+_logger = logging.getLogger(__name__)
+
+
+class ResCurrencyRateProvider(models.Model):
+    _name = 'res.currency.rate.provider'
+    _description = 'Currency Rates Provider'
+    _inherit = ['mail.thread']
+    _rec_name = 'name'
```
pedrobaeza: This is already the default value

pedrobaeza: I think what you want is to have `display_name` feature...

currency_rate_update/models/res_currency_rate_provider.py
```diff
@@ -0,0 +1,313 @@
...
+from odoo import models, fields, api, _
+from odoo.exceptions import UserError
+
+_logger = logging.getLogger(__name__)
+
+
+class ResCurrencyRateProvider(models.Model):
+    _name = 'res.currency.rate.provider'
+    _description = 'Currency Rates Provider'
+    _inherit = ['mail.thread']
```
pedrobaeza, line 19: I don't think we need this overhead here.

currency_rate_update/models/res_currency_rate_provider.py
```diff
@@ -0,0 +1,313 @@
...
+    _rec_name = 'name'
+    _order = 'name'
```
pedrobaeza, line 20: You can't order by a non stored field.

currency_rate_update/models/res_currency_rate_provider.py
```diff
@@ -0,0 +1,313 @@
...
+    service = fields.Selection(
+        string='Source Service',
+        selection=lambda self: self._selection_service(),
+        required=True,
+        track_visibility='onchange',
+    )
+    currency_ids = fields.Many2many(
+        string='Currencies',
+        comodel_name='res.currency',
+        column1='provider_id',
```
pedrobaeza, line 44: No need of specifying these columns unless there are several many2many tables

currency_rate_update/models/res_currency_rate_provider.py
```diff
@@ -0,0 +1,313 @@
...
+        column2='currency_id',
+        help='Currencies to be updated by this provider',
+        domain=lambda self: self._domain_currency_ids(),
```
pedrobaeza: Is this working properly? Why do you need then the onchange?

currency_rate_update/models/res_currency_rate_provider.py
```diff
@@ -0,0 +1,313 @@
...
+        required=True,
+        default=lambda self: self._default_company_id(),
+    )
+    active = fields.Boolean(
+        default=True,
+        track_visibility='onchange',
+    )
+    service = fields.Selection(
+        string='Source Service',
+        selection=lambda self: self._selection_service(),
```
pedrobaeza: No need of a selection method. Use `selection_add` instead on inherited models. This way, we can see in the model definition the available options.

currency_rate_update/models/res_currency_rate_provider.py
```diff
@@ -0,0 +1,313 @@
...
+        res = {
+            'domain': {
+                'currency_ids': self._domain_currency_ids(),
+            },
+        }
+
+        return res
+
+    @api.onchange('currency_ids')
+    def _onchange_currency_ids(self):
```
pedrobaeza: This is totally redundant

currency_rate_update/models/res_currency_rate_provider.py
```diff
@@ -0,0 +1,313 @@
...
+            ))[0][1]
+
+    @api.multi
+    @api.depends('active', 'interval_type', 'interval_number')
+    def _compute_update_schedule(self):
+        for provider in self:
+            if not provider.active:
+                provider.update_schedule = _('Inactive')
+                continue
+
```
pedrobaeza, line 107: Please avoid these useless line breaks inside methods, as PEP8 advises to

currency_rate_update/models/res_currency_rate_provider.py
```diff
@@ -0,0 +1,313 @@
...
+                'number': self.interval_number,
+                'type': list(filter(
+                    lambda x: x[0] == provider.interval_type,
+                    self._fields['interval_type'].selection
+                ))[0][1],
+            }
+
+    @api.multi
+    def _domain_currency_ids(self):
+        domain = [('active', '=', True)]
```
pedrobaeza: This domain is redundant

currency_rate_update/models/res_currency_rate_provider.py
```diff
@@ -0,0 +1,313 @@
...
+            if not data:
+                continue
+            if newest_only:
+                data = [max(
+                    data,
+                    key=lambda x: datetime.strptime(x[0], '%Y-%m-%d')
+                )]
+
+            for content_date, rates in data:
+                timestamp = datetime.strptime(content_date, '%Y-%m-%d')
```
pedrobaeza: use `fields.Datetime.from_string` always, but even better, don't return strings now in v12!

currency_rate_update/README.rst
```diff
@@ -1,67 +1,92 @@
...
+This module provides base for building exchange rates providers and bundles
+following built-in providers:
+
+ * **European Central Bank** (ported by Grzegorz Grzelak - OpenGLOBE.pl):
+   reference rates are based on the regular daily query procedure between
+   central banks within and outside the European System of Central Banks,
+   which normally takes place at 2:15 p.m. (14:15) ECB time. Source is in
+   EUR, for more details see `corresponding ECB page <https://www.ecb.europa.eu/stats/policy_and_exchange_rates/euro_reference_exchange_rates/html/index.en.html>`_.
+
+This module is compatible with ``currency_rate_inverted`` module provided by
```
pedrobaeza, line 37: Isn't inverted already supported here?

currency_rate_update/wizards/res_currency_rate_update_wizard.py
```diff
@@ -0,0 +1,34 @@
...
+        column2='provider_id',
+    )
+
+    @api.multi
+    def action_update(self):
+        self.ensure_one()
+
+        self.provider_ids._update(self.date_from, self.date_to)
+
+        return {'type': 'ir.actions.act_window_close'}
```
pedrobaeza, line 34: This is redundant

currency_rate_update/README.rst
```diff
@@ -1,67 +1,92 @@
...
-Roadmap:
+To update historical currency rates:
 
-* Google Finance.
-* Updated daily from Citibank N.A., source in EUR. Information may be delayed.
-  This is parsed from an HTML page, so it may be broken at anytime.
+# Go to *Invoicing > Configuration > Currency Rates Providers*
+# Select specific providers
+# Launch *Actions > Update Rates Wizard*
+# Configure date rate and click *Update*
```
pedrobaeza: s/date rate/date interval

currency_rate_update/models/res_currency_rate_provider_ECB.py
```diff
@@ -0,0 +1,85 @@
...
+
+from odoo import models, api, _
+from odoo.exceptions import UserError
+
+
+class ResCurrencyRateProviderECB(models.Model):
+    _inherit = 'res.currency.rate.provider'
+
+    @api.model
+    def _selection_service(self):
```
pedrobaeza: Use `selection_add`

pedrobaeza (changes requested):
- Code review made in below comments
- Other thing to do is to squash some adjacent commits from same author, as they are very bloated
- And the last thing about the refactoring made, I see correct to have new scheme, following Odoo paradigm (you have a similar scheme in `account_bank_statement_import`, so no need to revert it IMO.

currency_rate_update/models/res_currency_rate.py
alexey-pelykh: In most cases, currency rate is set once, and changing it for a specific date is a special event that should be logged, which would be useful to understand why values have changed. As far as I understand, initially it will not generate any emails, just change records.

pedrobaeza: But with this, the cron will also log an entry each day.

alexey-pelykh: for each _new_ day, to be precise

alexey-pelykh: Old days would be kept as is unless manually updated or requested to update.

currency_rate_update/models/res_currency_rate_provider.py
alexey-pelykh: Name is computed, just no `_rec_name` needed

currency_rate_update/models/res_currency_rate_provider.py
alexey-pelykh: Don't you think it's important to keep track of changes on provider? Source/currencies/etc?

currency_rate_update/models/res_currency_rate_provider.py
alexey-pelykh: These columns have non-default names, thus I had to specify those to avoid migration

currency_rate_update/models/res_currency_rate_provider.py
alexey-pelykh: To set the default domain? Or that's not needed?

pedrobaeza: This domain should be for all the lifetime, but it doesn't work properly unfortunately due to the ORM. Use this technique instead and remove the onchanges: https://github.com/OCA/bank-payment/pull/548

currency_rate_update/models/res_currency_rate_provider.py
pedrobaeza: Not too much in fact. This is transient information for the aim of having the rates.

currency_rate_update/models/res_currency_rate.py
pedrobaeza: I still think it's not needed. It doesn't add too much value. You already have the rates with a date reference...

currency_rate_update/models/res_currency_rate_provider.py
alexey-pelykh: Why?

currency_rate_update/models/res_currency_rate_provider.py
pedrobaeza: I would accept it if errors are logged in this chatter and are displayed in an easy way. Maybe you can put even an activity for an error.

currency_rate_update/models/res_currency_rate_provider.py
alexey-pelykh: Why to use full datetime?

currency_rate_update/models/res_currency_rate_provider.py
pedrobaeza: Because it's the domain that already applies ORM if you don't pass `active_test=False` in context.

currency_rate_update/wizards/res_currency_rate_update_wizard.py
alexey-pelykh: Why? It will autoclose?

currency_rate_update/models/res_currency_rate_provider.py
pedrobaeza: giphy

currency_rate_update/models/res_currency_rate_provider.py
alexey-pelykh: So remove tracking and add error logging? How to do the latter?

currency_rate_update/models/res_currency_rate.py
alexey-pelykh: Yet if it was used and then changed - it's better to keep at least some record of rate being altered

currency_rate_update/models/res_currency_rate.py
alexey-pelykh: If set manually - remove the reference to provider. Since user is the provider

currency_rate_update/models/res_currency_rate_provider.py
pedrobaeza: Sorry, I meant `fields.Date.from_string`. My point is to abstract `datetime` library. This can be beneficial for things like testing (https://twitter.com/PedroMBaeza/status/1069901067004141573)

currency_rate_update/models/res_currency_rate_provider.py
pedrobaeza: This can be done with manual calls to `post` method (not exact name) when an error happens. In previous versions, this was done in an ugly Text field if I don't remember bad.

currency_rate_update/models/res_currency_rate.py
pedrobaeza: But overhead each record with this inheritance?

currency_rate_update/models/res_currency_rate.py
pedrobaeza: OK, please put this as docstring in the method (and it's interesting to always do it on all overwrites that can require an explanation)

pedrobaeza: What happen if you change the date?

alexey-pelykh: Valid point

currency_rate_update/models/res_currency_rate.py
alexey-pelykh: How much this overhead would be?

pedrobaeza: Each time the record is accessed, a lot of checks are done for example. Check mail.thread code for getting you a bit scared.

currency_rate_update/models/res_currency_rate_provider.py
```diff
@@ -0,0 +1,299 @@
...
+        required=True,
+        track_visibility='onchange',
+    )
+    update_schedule = fields.Char(
+        string='Update Schedule',
+        compute='_compute_update_schedule',
+    )
+    next_run = fields.Date(
+        string='Next scheduled update',
+        default=fields.Date.today(),
```
kirca: You should just pass `fields.Date.today` to get as default "today's date" when the record is created. Currently you are calling it which will get evaluated when the field is instantiated instead.

currency_rate_update/README.rst
alexey-pelykh: The actual _usage_ of the inverted rate - no, it _stores_ the inverted rate if there's one and there's a module installed and this currency is inverted: https://github.com/brainbeanapps/currency/blob/e609c9d946ce93ee90385ad5e6ae6b5eb10bb4ea/currency_rate_update/models/res_currency_rate_provider.py#L248

currency_rate_update/wizards/res_currency_rate_update_wizard.py
alexey-pelykh: I don't think I've added this because of nothing - when the button is pressed, I need to close the dialog window.

currency_rate_update/models/res_currency_rate_provider.py
alexey-pelykh: Good catch, fixed!

alexey-pelykh:
@pedrobaeza regarding the commits squashing - agree, the history is huge, yet how to select which commits to squash?

currency_rate_update/wizards/res_currency_rate_update_wizard.py
pedrobaeza: But that's the default action. Have you tried without this?

pedrobaeza:
I always squash together by author and let the commit comments together. The only exception are migrations, that I don't squash together although the author is the same.

currency_rate_update/models/res_currency_rate.py
alexey-pelykh: I gave it some thought - seems still valid to me

currency_rate_update/models/res_currency_rate_provider.py
alexey-pelykh: Thanks!

currency_rate_update/models/res_currency_rate.py
alexey-pelykh: Otherwise it would be super-hard to track down the changes

alexey-pelykh:
@pedrobaeza I'll need your assistance with squashing the commits, I'm unsure about all cases

alexey-pelykh:
@pedrobaeza I've squashed some translation-related commits, for others I'm not certain

pedrobaeza:
Reduced a bit the commit history, although first commits are a hell due to constant merge operations.

pedrobaeza (approved):

MiquelRForgeFlow (approved):

currency_rate_update/models/res_currency_rate_provider.py
```diff
@@ -0,0 +1,295 @@
...
+    @api.multi
+    @api.depends('active', 'interval_type', 'interval_number')
+    def _compute_update_schedule(self):
+        for provider in self:
+            if not provider.active:
+                provider.update_schedule = _('Inactive')
+                continue
+
+            provider.update_schedule = _('%(number)s %(type)s') % {
+                'number': self.interval_number,
```
novawish: Should be `provider.interval_number`

novawish (changes requested):

currency_rate_update/models/res_currency_rate_provider.py
alexey-pelykh: Good catch!

alexey-pelykh:
@novawish please re-review

novawish (approved):

pedrobaeza:
Ouch, @alexey-pelykh you screw up my squashing!! Why? It took me a lot of time! I'll try to rescue it.

alexey-pelykh:
@pedrobaeza I'm so-so sorry, it totally slipped my mind :(

pedrobaeza:
OK, I have been able to recover it, I have amended your migration commit with the last comment, and forced push to branch 12.0. Please take attention with this next time.

alexey-pelykh:
@pedrobaeza super-sorry, I owe you a beer

PR #48, opened 2019-04-25
[12.0][IMP] currency_rate_update: allow instance access

alexey-pelykh:


currency_rate_update/models/res_currency_rate_provider.py
```diff
@@ -285,11 +285,16 @@ def _scheduled_update(self):
 
         _logger.info('Scheduled currency rates update complete.')
 
-    @api.model
-    def _get_supported_currencies(self, service):
+    @api.multi
+    def _get_supported_currencies(self):  # pragma: no cover
+        self.ensure_one()
+
```
pedrobaeza: Why adding this ugly empty lines inside a method?

currency_rate_update/models/res_currency_rate_provider.py
```diff
@@ -285,11 +285,16 @@ def _scheduled_update(self):
...
 
-    @api.model
-    def _obtain_rates(self, service, base_currency, currencies, date_from,
-                      date_to):
+    @api.multi
+    def _obtain_rates(
+        self, base_currency, currencies, date_from, date_to
+    ):  # pragma: no cover
+        self.ensure_one()
+
```
pedrobaeza: The same about the empty line

currency_rate_update/models/res_currency_rate_provider_ECB.py
```diff
@@ -19,9 +19,11 @@ class ResCurrencyRateProviderECB(models.Model):
         selection_add=[('ECB', 'European Central Bank')],
     )
 
-    @api.model
-    def _get_supported_currencies(self, service):
-        if service == 'ECB':
+    @api.multi
+    def _get_supported_currencies(self):
+        self.ensure_one()
```
pedrobaeza, line 24: No need to repeat the `self.ensure_one()`. Even more. This method doesn't need to ensure that.

currency_rate_update/models/res_currency_rate_provider_ECB.py
```diff
@@ -19,9 +19,11 @@ class ResCurrencyRateProviderECB(models.Model):
...
+
+        if self.service == 'ECB':
```
pedrobaeza: It's better for indentation to put `if self.service != 'ECB': return super...` and develop the rest of the lines with no indentation.

pedrobaeza (changes requested):

currency_rate_update/models/res_currency_rate_provider_ECB.py
alexey-pelykh: it's not repeating, super's `self.esnure_one()` may not be called

currency_rate_update/models/res_currency_rate_provider.py
alexey-pelykh: updated

currency_rate_update/models/res_currency_rate_provider.py
alexey-pelykh: updated

currency_rate_update/models/res_currency_rate_provider_ECB.py
pedrobaeza: OK, this is more a developer highlight than other thing, so it's not so important to be removed or not.

currency_rate_update/models/res_currency_rate_provider_ECB.py
alexey-pelykh: Done

currency_rate_update/models/res_currency_rate_provider_ECB.py
```diff
@@ -19,47 +19,51 @@ class ResCurrencyRateProviderECB(models.Model):
...
-                xml.sax.parse(response, handler)
-
-            return handler.content
-
-        return super()._obtain_rates(service, base_currency, currencies,
-                                     date_from, date_to)
+    @api.multi
+    def _get_supported_currencies(self):
+        self.ensure_one()
+
```
pedrobaeza: Still putting useless empty lines inside a method...

alexey-pelykh: done

currency_rate_update/models/res_currency_rate_provider_ECB.py
pedrobaeza: But I see more cases. I have put this one, but I mean all the empty lines inside a method.

alexey-pelykh: Well, logical separation?

pedrobaeza: We have talked about this other times:

* PEP8 encourages for not doing it.
* If you need a visual separation, better to put a comment line that also explains what the logical block is.
* This bloats the real number of lines/diff, and so on.

alexey-pelykh: But code like
```
a = something()
# here goes comment
b = something()
```
is less readable than
```
a = something()

# here goes comment
b = something()
```

If that's so strict, why it's not enforced via lint?

alexey-pelykh:
@pedrobaeza added comments and removed all empty lines that were not very useful to my understanding

pedrobaeza (approved):
Thanks for following my comments

tbaden (approved):
code review: lgtm

PR #50, opened 2019-04-30
[12.0][IMP] currency_rate_update: tests to succeed on non-banking days

alexey-pelykh:


alexey-pelykh:
@pedrobaeza right now tests are going to fail on non-banking days

pedrobaeza (approved):
OK, I think we can fast-track this

PR #52, opened 2019-05-04
[12.0][ADD] currency_rate_update_transferwise: TransferWise.com exchange rates

alexey-pelykh:
This module adds [TransferWise.com](https://transferwise.com/) currency exchange rates provider.


alexey-pelykh:
@mreficent @novawish is there a chance you're interested in this module?

alexey-pelykh:
@jbeficent @lreficent I'd like to request some fraction of your time to take a look at this module as PSC of the project

carlo-github:
> This module adds [TransferWise.com](https://transferwise.com/) currency exchange rates provider.

Thank you. I made some test with this module.. that's work as I need.. I have RON as main currency and I obtain the rates against EUR.. the only problem are the rate values that doesn't corrispond to the official rates.. Any know reason for that?

alexey-pelykh:
@carlo-github because it's not the official rates provider? :) These are rates that are used by TransferWise and some companies use those

carlo-github:
> @carlo-github because it's not the official rates provider? :) These are rates that are used by TransferWise and some companies use those

Thank you again..
So the best solution is OpenExchangeRates.org with a paid plan and I can have the RON as main currency or EUR or any other currencies? or do you have other suggest?

alexey-pelykh:
@carlo-github yes, if RON is main currency

alexey-pelykh:
@lreficent it would be great to hear your opinion on this

daramousk (approved):
There is currently the possibility of the user not to add any currencies in the currency rate provider, which in turn does not update anything. Not blocking but maybe this should have been handled differently (required field or something) else.

alexey-pelykh:
@daramousk thanks for review and the report! This issue is related to base module, from what I can see.

artemsenko (approved):

PR #56, opened 2019-05-27
[11.0][FIX] account_fx_spot: fix test 

LoisRForgeFlow:
Fix test  since now it is not possible to set the same currency in a journal as the company's one. (ref: https://github.com/odoo/odoo/commit/22f0da809e84618901b7ca4b78582abd64c03c55)

This fix does not need to be ported to v12.

cc @angelmoya @pedrobaeza 

@Eficent


angelmoya (approved):
:+1:

Issue #62, opened 2019-07-23
[12.0][BUG][currency_rate_update] Daily rates stop updating automatically on the end of weeks


gaikaz:
If configured to update daily, the `res.currency.rate.provider` gets stuck on the end of weeks. I have been seeing this on multiple databases and now found a way to reproduce it on OCA runbot.

**Steps to reproduce**
1. Set initial state:
  * Company currency set to `EUR`
  * Currency Rate Provider record:
    * Source Service: `European Central Bank`
    * Scheduled update interval: `1 Day(s)`
    * Last successful update: `07/01/2019` _(July 1)_
    * Next scheduled update: `07/02/2019`
    * Currencies: `USD`
  * Remove all (or at least relevant) rates from _Invoicing>Configuration>Accounting>Currencies>USD>Rates_
2. Imitate 5 daily scheduler (`Currency Rates Update (OCA) daily`) runs with `Run Manually` button by clicking it 5 times.

**Current behaviour**
First 4 runs will create currency rate records for dates `07/02/2019`, `07/03/2019`, `07/04/2019` and `07/05/2019`.
With each scheduler run, `Currency Rate Provider` record will update `Last/Next ... update` fields by increasing them by 1. _(which probably is an issue on it's own, as field named "Last successful update" doesn't make sense here)_.
This will repeat until 5th cron run, at which point the state becomes:
    * Last successful update: `07/05/2019`
    * Next scheduled update: `07/06/2019`
At which point any further runs of the scheduler will not do anything, even though there will not be any errors/warnings or messages about this in the log.

This can also be repeated with any other pair of initial dates. For example:
Initial state:
    * Last successful update: `07/06/2019`
    * Next scheduled update: `07/07/2019`
Gets stuck at state:
    * Last successful update: `07/12/2019`
    * Next scheduled update: `07/13/2019`

Initial state:
    * Last successful update: `07/16/2019`
    * Next scheduled update: `07/17/2019`
Gets stuck at state:
    * Last successful update: `07/19/2019`
    * Next scheduled update: `07/20/2019`

**Expected behaviour**
The rates should continue to be updated until current date is reached.

alexey-pelykh:
I'll take a look a this issue

gaikaz:
I have investigated this with a colleague today and it looks like ECB simply doesn't provide the rates for weekends.
Can be seen here: https://www.ecb.europa.eu/stats/eurofxref/eurofxref-hist-90d.xml
[...]

PR #63, opened 2019-08-07
[12.0][FIX] currency_rate_update: no data on weekends

alexey-pelykh:
Addresses https://github.com/OCA/currency/issues/62


alexey-pelykh:
@gaikaz please check the fix

gaikaz (commented):
Tested and works.
It would probably be a good idea to also add a "Known issues / Roadmap" section to `README.rst`, mentioning that ECB doesn't provide currency rates for weekends and it is currently expected behaviour.

alexey-pelykh:
@pedrobaeza fast-track or not qualified?
@gaikaz well, I don't think it's worth mentioning that banks are closed on non-banking days, so don't expect European Central Bank to be working on official holidays and non-banking days :)

pedrobaeza (approved):

Issue #60, opened 2019-07-06
403 forbidden


carlo-github:
HTTP Error 403: Forbidden odoo12
when perform action "Update Rates Wizard" in     Currency Rates Providers/OpenExchangeRates.org
currency EUR RON

ihave put de app id in configuration

![Firefox_Screenshot_2019-07-06T13-43-28 056Z](https://user-images.githubusercontent.com/36795969/60757035-8b11f000-a005-11e9-99c6-c49c53451916.png)
![Firefox_Screenshot_2019-07-06T13-44-04 885Z](https://user-images.githubusercontent.com/36795969/60757036-8b11f000-a005-11e9-80fe-7a504fea4930.png)
![Firefox_Screenshot_2019-07-06T13-44-41 905Z](https://user-images.githubusercontent.com/36795969/60757037-8b11f000-a005-11e9-92fe-a66b1546e492.png)

gaikaz:
Getting same error. After investigation, the problem seems to be rising from incorrect authorization header being sent to OXR API.

Suspected line of code:
https://github.com/OCA/currency/blob/12.0/currency_rate_update_oxr/models/res_currency_rate_provider_OXR.py#L106

On this line, the header is set like so:
`'Bearer %s' % self.company_id.openexchangerates_app_id`
Which simplified would be:
`'Bearer YOUR_APP_ID'`
Which by urllib is most likely being parsed to:
`"Authorization: Bearer YOUR_APP_ID"`

According to OXR API documentation, correct header would be:
`"Authorization: Token YOUR_APP_ID"`

OXR API documentation state:

> **Please note:** The format of the HTTP header must be exactly as above ...

Documentation reference:
https://docs.openexchangerates.org/docs/authentication#http-header-authentication

I also tested this with Postman, which confirms, that OXR returns `403` if called with authorization header with keyword `Bearer` and gives correct response with keyword `Token`.

carlo-github:
> Getting same error. After investigation, the problem seems to be rising from incorrect authorization header being sent to OXR API.
> 
> Suspected line of code:
> https://github.com/OCA/currency/blob/12.0/currency_rate_update_oxr/models/res_currency_rate_provider_OXR.py#L106
> 
> On this line, the header is set like so:
> `'Bearer %s' % self.company_id.openexchangerates_app_id`
> Which simplified would be:
> `'Bearer YOUR_APP_ID'`
> Which by urllib is most likely being parsed to:
[...]

PR #64, opened 2019-08-09
[12.0][FIX] currency_rate_update_oxr: use Token instead of Bearer

alexey-pelykh:
As per https://github.com/OCA/currency/issues/60#issuecomment-518991006 @gaikaz

### DIFF
--- a/currency_rate_update_oxr/models/res_currency_rate_provider_OXR.py
+++ b/currency_rate_update_oxr/models/res_currency_rate_provider_OXR.py
@@ -103,6 +103,6 @@ class ResCurrencyRateProviderOXR(models.Model):
         request = urllib.request.Request(url)
         request.add_header(
             'Authorization',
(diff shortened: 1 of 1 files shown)


gaikaz:
Ok. I didn't fully understand, how the issue got solved that day. Now I see it. I had EUR as base currency too. 🤦‍♂️

Now about this header, is there even a case, where the url parameter `app_id` is missing and header is needed? Does is get parsed out somehow?

alexey-pelykh:
Nope, both Header and `app_id` query parameter is passed :)

alexey-pelykh:
@lreficent any feedback on this?
@gaikaz please review this PR if possible

gaikaz (approved):
Can't test it functionally right now, but the change LGTM

Though, as I understand, this part of code is just a redundancy for adding the `app_id` if OXR starts prioritizing the http header instead of URL parameter `app_id`, how it is now.

PR #68, opened 2019-09-25
[12.0][MIG] account_fx_spot

MiquelRForgeFlow:
Standard migration.


bosd:
@MiquelRForgeFlow What is the status of this PR?

MiquelRForgeFlow:
I suppose that is in "to be reviewed" status.

bosd:
Got an server error on a functional test, when registering an payment.
Register Payment error

MiquelRForgeFlow:
Sorry, we don't need this anymore, so I will not dedicate more time to this for the time being.

PR #71, opened 2019-10-18
[12.0][IMP] currency_rate_update for non EUR companies

### LINKED ISSUE #70: [12.] module currency_rate_update European Central Bank Update for non EURO companies
Hello all
In Version 12 ECB currency update can only be used if the companies base currency is EUR.
Is there a reasoning for this change from previous versions?

I understand the warning message below but why the removal of the functionality?
('European Central Bank is suitable only for companies with EUR as base currency!', '')

Regards,

Martronic-SA:
That PR reintroduces the v10 and v11 behavior by letting companies to use European Central Bank for other currencies than EUR.
If another currency is used as base currency, an inverted calcuation on rates is done.
This might require more investigation on the accuracy of the calculation for foreign currencies between them.
This should fix https://github.com/OCA/currency/issues/70 but discussion on introducing inverted calculation in core is still open.

### DIFF
--- a/currency_rate_update/__manifest__.py
+++ b/currency_rate_update/__manifest__.py
@@ -4,7 +4,7 @@
 
 {
     'name': 'Currency Rate Update',
-    'version': '12.0.1.1.2',
+    'version': '12.0.1.2.0',
     'author':
         'Camptocamp, '
         'Brainbean Apps, '
--- a/currency_rate_update/models/res_currency_rate_provider_ECB.py
+++ b/currency_rate_update/models/res_currency_rate_provider_ECB.py
@@ -8,8 +8,7 @@ from datetime import date, timedelta
 from urllib.request import urlopen
 import xml.sax
 
-from odoo import models, fields, api, _
-from odoo.exceptions import UserError
+from odoo import models, fields, api
 
 
 class ResCurrencyRateProviderECB(models.Model):
@@ -34,7 +33,7 @@ class ResCurrencyRateProviderECB(models.Model):
                 'SIT', 'SKK', 'CHF', 'ISK', 'NOK', 'HRK', 'RUB', 'TRL',
                 'TRY', 'AUD', 'BRL', 'CAD', 'CNY', 'HKD', 'IDR', 'ILS',
                 'INR', 'KRW', 'MXN', 'MYR', 'NZD', 'PHP', 'SGD', 'THB',
-                'ZAR',
+                'ZAR', 'EUR'
             ]
 
     @api.multi
@@ -43,13 +42,11 @@ class ResCurrencyRateProviderECB(models.Model):
         if self.service != 'ECB':
             return super()._obtain_rates(base_currency, currencies, date_from,
                                          date_to)  # pragma: no cover
-
-        # This provider only serves EUR-to-??? exchange rates
-        if base_currency != 'EUR':  # pragma: no cover
-            raise UserError(_(
-                'European Central Bank is suitable only for companies'
-                ' with EUR as base currency!'
-            ))
+        invert_calculation = False
+        if base_currency != 'EUR':
+            invert_calculation = True
+            if base_currency not in currencies:
+                currencies.append(base_currency)
 
         # Depending on the date range, different URLs are used
         url = 'https://www.ecb.europa.eu/stats/eurofxref'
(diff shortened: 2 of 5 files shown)


currency_rate_update/models/res_currency_rate_provider_ECB.py
```diff
@@ -44,12 +44,11 @@ def _obtain_rates(self, base_currency, currencies, date_from, date_to):
...
                                          date_to)  # pragma: no cover
 
-        # This provider only serves EUR-to-??? exchange rates
-        if base_currency != 'EUR':  # pragma: no cover
-            raise UserError(_(
-                'European Central Bank is suitable only for companies'
-                ' with EUR as base currency!'
-            ))
+        if base_currency != 'EUR' and \
+                base_currency in self._get_supported_currencies():
```
alexey-pelykh: Since you've added EUR to supported currencies list, you can just check for `base_currency != 'EUR'`

currency_rate_update/__manifest__.py
```diff
@@ -4,7 +4,7 @@
 
 {
     'name': 'Currency Rate Update',
-    'version': '12.0.1.1.2',
```
alexey-pelykh, line 7: This is not a patch release, it's a full-scale minor change.

currency_rate_update/views/res_currency_rate_provider.xml
```diff
@@ -44,6 +44,9 @@
                     <group name="options">
                         <group>
                             <field name="service"/>
+                            <div attrs="{'invisible':[('service','!=','ECB')]}" class="text-muted" colspan="2">
+                              European Central Bank exchange rate service is suited for companies using EUR as main currency.
```
alexey-pelykh: I presume that we don't need this note?

alexey-pelykh (changes requested):

alexey-pelykh:
@Martronic-SA please fix Travis issues

currency_rate_update/__manifest__.py
Martronic-SA: 12.0.1.2.1 ?

alexey-pelykh: `12.0.1.2.0`, you can even omit this change - OCA bot will take care of it during merge

alexey-pelykh: Why .1 in the end?

alexey-pelykh:
2 more issues in linter :)

currency_rate_update/views/res_currency_rate_provider.xml
Martronic-SA: Finally I think it'simportant for the user to know that currencies are converted base on EUR and that the currency exchange rate between foreign currencies might differ from real currencies exchange rates.

alexey-pelykh: Yet we're not doing double conversion, and since ECB provides 1 rate only, I assume it's bidirectional? Yet telling that may not be an issue in the end

alexey-pelykh:
@Martronic-SA awesome! Please cover this usecase with a test case and squash the commits as well

pedrobaeza:
I have to say that allowing this will give the false sensation that the data is accurate, which is not the case. I'm not very in favor of having it, or not without enough warnings.

alexey-pelykh:
Yes, that was the primary reason of dropping that support in v12

Martronic-SA:
> 
> 
> I have to say that allowing this will give the false sensation that the data is accurate, which is not the case. I'm not very in favor of having it, or not without enough warnings.

I've added the following comment in the form:

pedrobaeza:
s/Your are/You are, and better to use alert/warning BS class for highlighting it.

alexey-pelykh (approved):

alexey-pelykh (commented):
.

pjrobertson:
Just come across this issue myself with the currency_rate_update plugin. Our base currency is also not EUR. Thanks @Martronic-SA for putting in the hard work to add this feature back into v12.0 plugin. I look forward to it getting merged into master!

MiquelRForgeFlow (approved):

alexey-pelykh (changes requested):
Please squash the commits into one, there's no sense in having 7 commits for this change

Martronic-SA:
@alexey-pelykh I hope this is ok now! :)

alexey-pelykh (approved):

MiquelRForgeFlow:
no need to change number version at merging

alexey-pelykh:
@MiquelRForgeFlow that's optional, right? non-blocking that is

MiquelRForgeFlow:
I didn't mean to change the commit. I mean that the person who does "oca merge" doesn't have to put path/minor/manor because the commit already includes the version number update.

alexey-pelykh:
@MiquelRForgeFlow ah, got what you mean, yes, that's logical :)

pjrobertson:
Great to see this! Thanks all for your hard work :)

PR #80, opened 2020-07-03
[13.0][FIX] currency_rate_update: Migration to 13.0

alexey-pelykh:


currency_rate_update/models/res_currency_rate_provider.py
```diff
@@ -195,7 +195,7 @@ def _process_rate(self, currency, rate):
 
         Module = self.env["ir.module.module"]
 
-        currency_rate_inverted = Module.sudo().search(
```
pedrobaeza: Why? Previous code was ok

pedrobaeza (changes requested):
What has happened?

currency_rate_update/models/res_currency_rate_provider.py
alexey-pelykh: https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-13.0#technical-method-to-migrate-a-module-from-120-to-130-branch sudo() is deprecated?

alexey-pelykh:
@pedrobaeza migration seems to have been corrupted/incomplete at some point, and that needs fixing.

currency_rate_update/models/res_currency_rate_provider.py
pedrobaeza: sudo(user) is, not sudo(). Is documentation not clear enough?

alexey-pelykh: Oh! Actually, I assumed that entire method was deprecated

alexey-pelykh: Updated

pedrobaeza (approved):

Issue #84, opened 2020-09-18
[Question] currency_rate_update only sync with USD? At least not working with GBP, JPY, DKK


rafaelbn:
Hello,

I'm trying to sync USD and GBP, USD works but GBP not.

Any idea?

Thanks in advance
Regards

alexey-pelykh:
@rafaelbn a bit more details would be much appreciated, since it works for me for EUR/USD/AUD/GBP/UAH with EUR as base currency

rafaelbn:
Hello @alexey-pelykh , thanks for answering.

Well we should then look deeper... It just that a normal configuration doesn't work as expected

![2020-09-18_12-01-36](https://user-images.githubusercontent.com/8736623/93585470-25f6b000-f9a7-11ea-81dd-1ff8c8e09eb5.png)

alexey-pelykh:
Is that reproducible on runbot? I'm not using ECB on my instance thus can not verify its behaviour over long period of time

victoralmau:
Related with this issue, i have created https://github.com/OCA/currency/pull/85 that pretend to solved this "special situation" and describe the problem.

rafaelbn:
Closing then, we continue ni https://github.com/OCA/currency/pull/85

PR #85, opened 2020-09-22
[FIX] currency_rate_update: Prevent last_successful_run update if all…

victoralmau:
Prevent last_successful_run update if all records not create to every currencies and dates needed.

**Example**: BCE update rate in 14:00+01:00 and if we define ir_cron to 02:00 hour, read all the info but not exists records from some currencies from today but updated last_successful_run field and NOT save any records to some currencies and dates. Tomorrow the same problem and without any record save about currency rate.

**Solution**: Add method `_get_close_date` that allow define close hour. If defined, only when hour is greater is executed update for this provider.

**Issue related**: https://github.com/OCA/currency/issues/84

Please, @Yajo and @pedrobaeza   review it

@Tecnativa TT25624

### DIFF
--- a/currency_rate_update/models/res_currency_rate_provider.py
+++ b/currency_rate_update/models/res_currency_rate_provider.py
@@ -1,6 +1,7 @@
 # Copyright 2009-2016 Camptocamp
 # Copyright 2010 Akretion
 # Copyright 2019-2020 Brainbean Apps (https://brainbeanapps.com)
+# Copyright 2020 Tecnativa - Víctor Martínez
 # License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html).
 
 import logging
@@ -96,6 +97,10 @@ class ResCurrencyRateProvider(models.Model):
             'Scheduled update interval must be greater than zero!'
         )
     ]
+
+    @api.model
+    def _get_close_date(self):
+        return False
 
     @api.multi
     @api.depends('service')
@@ -318,7 +323,13 @@ class ResCurrencyRateProvider(models.Model):
                     provider.next_run - provider._get_next_run_period()
                 )
                 date_to = provider.next_run
-                provider._update(date_from, date_to, newest_only=True)
+                if (date_to != fields.Date.today()) or (
+                    date_to == fields.Date.today() and (
+                        not provider._get_close_time()
+                        or datetime.now().hour >= provider._get_close_time()
+                    )
+                ):
+                    provider._update(date_from, date_to, newest_only=True)
 
         _logger.info('Scheduled currency rates update complete.')
 
--- a/currency_rate_update/models/res_currency_rate_provider_ECB.py
+++ b/currency_rate_update/models/res_currency_rate_provider_ECB.py
@@ -18,6 +18,12 @@ class ResCurrencyRateProviderECB(models.Model):
         selection_add=[('ECB', 'European Central Bank')],
     )
 
+    @api.model
+    def _get_close_time(self):
+        if self.service == 'ECB':
+            return 17
+        return super()._get_close_time()
+
     @api.multi
     def _get_supported_currencies(self):
         self.ensure_one()
--- a/currency_rate_update/data/cron.xml
+++ b/currency_rate_update/data/cron.xml
@@ -3,12 +3,12 @@
     Copyright 2019 Brainbean Apps (https://brainbeanapps.com)
     License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
 -->
-<odoo>
+<odoo noupdate="1">
 
     <record model="ir.cron" id="ir_cron_currency_rates_update_every_day">
         <field name="name">Currency Rates Update (OCA) daily</field>
-        <field name="interval_number">1</field>
-        <field name="interval_type">days</field>
+        <field name="interval_number">12</field>
+        <field name="interval_type">hours</field>
         <field name="numbercall">-1</field>
         <field name="state">code</field>
         <field name="nextcall" >2019-01-01 00:01:00</field>
--- a/currency_rate_update/readme/CONTRIBUTORS.rst
+++ b/currency_rate_update/readme/CONTRIBUTORS.rst
@@ -20,3 +20,9 @@
 * Dmytro Katyukha <firemage.dima@gmail.com>
 * Jesús Ventosinos Mayor <jesus@comunitea.com>
 * Alexey Pelykh <alexey.pelykh@brainbeanapps.com>
+
+* `Tecnativa <https://www.tecnativa.com>`_:",
+
+  * Víctor Martínez
+  * Pedro M. Baeza
+


alexey-pelykh:
I'm not sure how it will react for weekends, since some providers don't provide data for those days. And we should not fill those with fake data then, since Odoo will use the last known. I think it would be better to do this force-fail at a provider level, not at base module level. In other words, if provider knows that it provides all and everything, it could fail if something is missing - thus rescheduling itself

victoralmau:
> I'm not sure how it will react for weekends, since some providers don't provide data for those days. And we should not fill those with fake data then, since Odoo will use the last known. I think it would be better to do this force-fail at a provider level, not at base module level. In other words, if provider knows that it provides all and everything, it could fail if something is missing - thus rescheduling itself

Ok, i understand the problem about "closed dates" and other days (https://www.ecb.europa.eu/home/contacts/working-hours/html/index.en.html) when forex market are closed.
This situation happen to any provider (because it'is possible to exec cron and call provider before closed forex market and NOT exist yet rate to some currencies).

I proposse update this changed (related to this problem):

- When not data received from provider, check if any date (between date_from and date_to) are today, in this case NOT exec _schedule_next_run (becasue probably data not exist yet related to close date in forex marked).

- If data received, check if any date (between date_from and date_to) are today, in this case, search if exists all the records that we expeceted (number of records in res_currency_rate filter by today date is equal to currencies related to this provider) is correct, in this case OK, update _schedule_next_run.

If you think that this proposed "solved" the problem about today, i change it now and review later.

On ther other hand, if we add this info in the documentation will help anybody with similar problems.

This info about close session in forex marked related to regions is very usefull:

main-qimg-3b74f3b2c5221017ff4202b46ed585b2-600x376

pedrobaeza:
What about associate a closing date to each provider (ECB at 17:00 for example - it can be hardcoded), and then check if the cron is running before that hour and then don't retrieve data at all if the day to fetch rate is today? If no closing date is defined, then work as always. 

Anyway, the main problem right now is the cron not being noupdate="1" for being able to change the hour for workarounding the problem. Maybe you can put only that change for fast-tracking it.

victoralmau:
This option can solve the problem but IMO this complicate addon logic and specially configuration (because need to know close hour to every forex market) and too keep on existing the problem that comment previously.

Maybe, my last comment that suggest only NOT update a field in provider if date of filters is today can simplified the "problem" and not blocking the process.

Is important too add more documentation about it and of course, change noupdate option in cron data.

pedrobaeza:
Well, it's not configuration, as it's hardcoded in the provider, so the developer must include it when adding a provider, and for user it's the same. What do you think about this, @alexey-pelykh ?

currency_rate_update/models/res_currency_rate_provider.py
```diff
@@ -176,7 +177,7 @@ def _update(self, date_from, date_to, newest_only=False):
 
             if not data:
                 if is_scheduled:
-                    provider._schedule_next_run()
+                    _logger.warning('Currency Rate Provider "%s" withoud data' % provider.name)
```
yajo:
```suggestion
                    _logger.warning('Currency Rate Provider "%s" without data', provider.name)
```

currency_rate_update/models/res_currency_rate_provider.py
```diff
@@ -227,7 +228,30 @@ def _update(self, date_from, date_to, newest_only=False):
                         })
 
             if is_scheduled:
-                provider._schedule_next_run()
+                # Search if every records (all dates and all currencies) are save
```
yajo:
```suggestion
                # Search if every records (all dates and all currencies) are saved
```

currency_rate_update/models/res_currency_rate_provider.py
```diff
@@ -227,7 +228,30 @@ def _update(self, date_from, date_to, newest_only=False):
...
+                    provider._schedule_next_run()
+                else:
+                    _logger.info(
+                        "Missing records between dates %s and %s from currencies: %s | %s record(s) of %s" % (
+                        date_from,
+                        date_to,
+                        provider.currency_ids.mapped('name'),
+                        total_records,
+                        total_correct
+                    ))
```
yajo:
```suggestion
                        "Missing records between dates %s and %s from currencies: %s | %s record(s) of %s",
                        date_from,
                        date_to,
                        provider.currency_ids.mapped('name'),
                        total_records,
                        total_correct
                    )
```

currency_rate_update/models/res_currency_rate_provider.py
```diff
@@ -227,7 +228,30 @@ def _update(self, date_from, date_to, newest_only=False):
...
+                total_records = CurrencyRate.search_count([
+                    ('company_id', '=', provider.company_id.id),
+                    ('currency_id', 'in', provider.currency_ids.ids),
+                    ('name', 'in', dates_need_check)
+                ])
+                total_correct = len(provider.currency_ids.ids) * len(dates_need_check)
+                if total_records == total_correct:
+                    provider._schedule_next_run()
+                else:
+                    _logger.info(
```
yajo: Maybe this should be a warning?

```suggestion
                    _logger.warning(
```

yajo (changes requested):
Some general coding comments:

yajo:
I think the problem is the data model of this module. 

First of all, if the time of the day when data should be fetched is important, then these fields should be `Datetime` and not `Date`, so that i.e. BCE can be rescheduled to run after 14:15+01:00 https://github.com/OCA/currency/blob/d1165a264cb23e700e257a2fcd57277c52edd1f2/currency_rate_update/models/res_currency_rate_provider.py#L78-L85

However, I'm not sure that those fields are really required, because:

1. `last_successful_run` can be obtained by this pseudocode: `min(self.currency_ids.mapped('date'))`
1. `next_run` is not needed because we can consider that the process must run as long as `last_successful_run <= date.today() - timedelta(days=1)`

Finally, once that is done, if you run the cron every hour, you'd get currency rates quicker without configuration, and providers that are done wouldn't be executed, so it wouldn't be a big problem too.

Summing all of that, you'd get an always-working solution out of the box.

alexey-pelykh:
My opinion is quite straightforward on this:

- Provider is responsible for providing data
- The data provided must be correct and consistent and that's responsibility of provider
- If for one reason or another provider can not provide data until some time in the day (e.g. 17:00), it should provide none.
- If for one reason or another provider was requested data for USD and CHF but provided only USD - it means that this higher-level module understands that as "ok, there's data for USD and no data for CHF and since provider did say that, that is truth"

So I'm completely with @pedrobaeza on this matter. In short: fixing module to accommodate ECB specifics is bad.

alexey-pelykh:
@Yajo to your comment:

However, I'm not sure that those fields are really required, because:

> last_successful_run can be obtained by this pseudocode: min(self.currency_ids.mapped('date'))

That is simply not true. Lack of data does not mean run was not successful. There's no data on weekends, why that run was not successful?

pedrobaeza:
And there are holidays that can extend the no-data period for more than 2 days.

alexey-pelykh:
As far as I can see, first step is to add `noupdate` on cron, that's clearly an issue in code. Next step is to fix ECB implementation to check data integrity and possibly ignore every request prior to 17:00 as that's ECB specific.

victoralmau:
With all this comments about it i think that maybe not exist an error but yes a bad configuration.
If somebody use BCE (or other provider) it's important to know hour when provider return info from today about a currency.

If only use BCE with EUR and USD, i configure provider with daily frequency, it's important for me to define cron to correct hour (> min hour that provider return info about my currencies from today) > This **not an error in addon BUT yes need to configure it correctly** for this use case. (With other provider is exactly the same).

If use BCE (or other provider) with: EUR, USD and CNY, i configure provider with daily frequency, it's importar form me to define cron to correct hour (> min hour that provider return info about **ALL my currencies** with most attention in this case CNY that probably "close" forex market later)

Too, it's important to know about it and this functionality is correct, if not received data is because not data for this day (forex market is closed) or not data YET (BAD configuration).

What is your opinion about it?

It's important to know the final decision to change some things in addon if really need or only `noupdate` in cron.

alexey-pelykh:
> Too, it's important to know about it and this functionality is correct, if not received data is because not data for this day (forex marked is closed) or not data YET (BAD configuration).

Perfect summary, I'd say. So, if no-data case happens and provider gives no error, it's a "forex marked is closed". Otherwise, if provider knows its limits, it should return error to notify of bad configuration

yajo:
@alexey-pelykh thanks for your comments.

According to what you said, I understand that this is true and the schedule fields needs to be `Datetime`, right?:

> these fields should be `Datetime` and not `Date`, so that i.e. BCE can be rescheduled to run after 14:15+01:00 https://github.com/OCA/currency/blob/d1165a264cb23e700e257a2fcd57277c52edd1f2/currency_rate_update/models/res_currency_rate_provider.py#L78-L85

I also see a bunch of hours being mentioned across the PR, and I want to take the chance to clarify (also to myself) that [**BCE releases currency exchange rates at around 16:00 CET**, as you can see in their website](https://www.ecb.europa.eu/stats/policy_and_exchange_rates/euro_reference_exchange_rates/html/index.en.html):

imagen

pedrobaeza:
I think the module must be resilient no matter the configuration of the cron (as well as if you want to put an hourly schedule), so I still see as best solution to hardcode that "closing hour" on BCE (but we can put a general closing hour mechanism, only hardcoding on each provider the exact hour).

Putting time in next schedule and last schedule is useless IMO.

victoralmau:
Add method `_get_close_date` that allow define close hour. If defined, only when hour is greater is executed update for this provider.

currency_rate_update/models/res_currency_rate_provider.py
```diff
@@ -138,6 +143,7 @@ def _compute_available_currency_ids(self):
 
     @api.multi
     def _update(self, date_from, date_to, newest_only=False):
+
```
pedrobaeza: Don't add extra empty lines inside method.

currency_rate_update/models/res_currency_rate_provider.py
```diff
@@ -318,7 +324,9 @@ def _scheduled_update(self):
                     provider.next_run - provider._get_next_run_period()
                 )
                 date_to = provider.next_run
-                provider._update(date_from, date_to, newest_only=True)
+                close_date = provider._get_close_date()
+                if not close_date or (close_date and datetime.now().hour >= close_date):
```
pedrobaeza: You must compare the close_date only for fetching data from today, not from any date interval.

pedrobaeza (changes requested):

pedrobaeza:
Check also Travis status

currency_rate_update/models/res_currency_rate_provider.py
victoralmau: > Check also Travis status

Travis solved

victoralmau:
> Check also Travis status

Travis solved

currency_rate_update/models/res_currency_rate_provider.py
```diff
@@ -97,6 +98,10 @@ class ResCurrencyRateProvider(models.Model):
         )
     ]
 
+    @api.model
+    def _get_close_date(self):
+        return False
```
yajo, line 103: Hmm I'm very sorry but this is the best way to have further problems.

Look, if you return a "time" here... it's OK in some time zone, but not OK in others. Also, if you split the date and time fields, you'll find yourself in weird corner cases, like somebody in Australia wanting to update currencies from BCE once a day, but it will never get data because when he starts the cron, it will be tomorrow in Australia for Europe.

If we really care about dates and times... well, then that's why you have datetime fields.

I'm sorry but I have to insist: `next_run` field must be of type `Datetime` or problems will never end. :shrug:

yajo (changes requested):

currency_rate_update/models/res_currency_rate_provider.py
pedrobaeza: @Yajo there's no TZ here. You always work with UTC and this time is in UTC and it must be processed that way. Your solution blows away the current scheme. Please restrict your review to possible problems to this approach (maybe the comparison is not done in UTC, but it should), but don't come again to change the approach.

yajo: Yes, I'm telling possible problems. :grin: The data model is wrong and must change...

You can, however, add a new Datetime field (ex. `next_run_dt`) and provide a fallback for other addons using a computed field that just returns the date part in `next_run`.

The problem is: what is, exactly, `2020-01-01`?

- For you, it's `2020-01-01 00:00:00 UTC+02:00`.
- For somebody in Fiji it is `2020-01-01 00:00:00 UTC+12:00`.
- For somebody in Samoa, it is `2020-01-01 00:00:00 UTC-11:00`.

So, when `datetime.today()` returns `2020-01-01` for the BCE, it might return a different date elsewhere. Thus, **there is an implicit TZ** in a date value.

Then, if you convert `2020-01-01` blidly into `2020-01-01 17:00:00 UTC`, it could happen that you are referring to `17:00 UTC` for today, for tomorrow or for yesterday, because the 1st half of that datetime value (`2020-01-01`) is timzeone-aware implicitly.

To make it even more fun, Odoo stores data as UTC, but displays it localized to the user, so an Odoo instance for a multinational company could display different execution dates (even if it happened to say the same time) on different timezones, both for this model and for `ir.cron`. You might be obtaining currency rates for yesterday if you're in Samoa and run the wizard, for example.

The fact that a date and time fields alone lack timezone info makes them unreliable data types when comparing to an absolute time.

So, in summary, if we don't start using a datetime field, we'll just be switching one problem for another.

alexey-pelykh: I guess `next_run` to be datetime is proper way to do things, as well as probably making cron to be run every hour. Yet in any case Odoo itself can have only 1 rate per date, and that date is w/o timezone, which kind-a makes it complex.

currency_rate_update/models/res_currency_rate_provider.py
```diff
@@ -318,7 +323,13 @@ def _scheduled_update(self):
                     provider.next_run - provider._get_next_run_period()
                 )
                 date_to = provider.next_run
-                provider._update(date_from, date_to, newest_only=True)
+                if (date_to != fields.Date.today()) or (
```
alexey-pelykh, line 326: This is some voodoo. If provider is requested data that it does not have, it should return none. I'm not sure I follow why higher-level module should consider close date of provider. Some providers don't even have one and have rate for every hour

currency_rate_update/models/res_currency_rate_provider_ECB.py
```diff
@@ -18,6 +18,12 @@ class ResCurrencyRateProviderECB(models.Model):
         selection_add=[('ECB', 'European Central Bank')],
     )
 
+    @api.model
+    def _get_close_date(self):
+        if self.service == 'ECB':
+            return 17
```
alexey-pelykh, line 24: Close date 17? :)

pedrobaeza: It should be `_get_close_time`?

currency_rate_update/models/res_currency_rate_provider.py
victoralmau: This is related to https://github.com/OCA/currency/pull/85#issuecomment-697223174

currency_rate_update/models/res_currency_rate_provider_ECB.py
victoralmau: BCE Provider define 16:00 (https://www.ecb.europa.eu/stats/policy_and_exchange_rates/euro_reference_exchange_rates/html/index.en.html) but us define 17:00 (to ensure that data already exists)

alexey-pelykh: At least yes, and how to do you propose to use this method for providers like XE.com or OXR?

victoralmau:
Superseed by: https://github.com/OCA/currency/pull/117

PR #88, opened 2020-10-09
[14.0][MIG] currency_rate_update: Migration to 14.0

feketemihai:


rvalyi (approved):

ventor-dev:
@pedrobaeza , should we merge it?

pedrobaeza:
There aren't enough reviews. Check https://github.com/OCA/odoo-community.org/blob/master/website/Contribution/CONTRIBUTING.rst#review

currency_rate_update/views/res_currency_rate_provider.xml
```diff
@@ -0,0 +1,149 @@
...
+                <sheet>
+                    <div class="oe_button_box" name="button_box">
+                        <button
+                            class="oe_stat_button"
+                            type="object"
+                            name="toggle_active"
+                            icon="fa-archive"
+                            groups="base.group_system"
+                        >
+                            <field
```
jcdrubay: In v14, this is not anymore the standard way, but anyway it works.

I think it should be something like this
```
                    <div class="oe_button_box" name="button_box"/>
                    <widget name="web_ribbon" title="Archived" bg_color="bg-danger" attrs="{'invisible': [('active', '=', True)]}"/>
```

jcdrubay (approved):
LGTM, only a tiny comment for the button Archive which could be transformed into v14 style.

currency_rate_update/views/res_currency_rate_provider.xml
feketemihai: @jcdrubay Fixed it in https://github.com/OCA/currency/pull/88/commits/0b3070d5666c88d5faccf3cc0af617d0cc6684cb. Thanks for the review.

pedrobaeza:
In runbot we have:
```
2020-12-01 06:36:45,020 160 WARNING openerp_test odoo.fields: Field res.currency.rate.rate: unknown parameter 'track_visibility', if this is an actual parameter you may want to override the method _valid_field_parameter on the relevant model in order to allow it
2020-12-01 06:36:45,020 160 WARNING openerp_test odoo.fields: Field res.currency.rate.provider_id: unknown parameter 'track_visibility', if this is an actual parameter you may want to override the method _valid_field_parameter on the relevant model in order to allow it
```
and please squash migration commits into one.

PR #89, opened 2020-10-11
[MIG] currency_rate_update: Migration to 14.0

tupaq:


pedrobaeza:
There's already another migration in #88, and you haven't preserved commit history. Please review the other instead of opening a new one, and take into account for other times the migration guide:

https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-14.0

PR #96, opened 2021-02-24
[12.0] ADD currency_rate_update_boi - Currency Rate Update: Bank of Italy

eLBati:


marcelsavegnago:
Helo @eLBati. If this is specific to Italy, wouldn't it be better to put it in the l10n-italy repository? Not that I have any restrictions but the question is more to understand the logic since in Brazil we put the module inside the l10n-brasil repository but I don't see that any team restriction to move the module to the currency repository.

https://github.com/OCA/l10n-brazil/blob/12.0/l10n_br_currency_rate_update/__manifest__.py

cc @renatonlima @rvalyi @mileo

eLBati:
@marcelsavegnago I think you are right.
Moving to https://github.com/OCA/l10n-italy/pull/2129

PR #102, opened 2021-03-27
[ADD] vietcombank currency rate

lehoangan:


alexey-pelykh:
@lehoangan many thanks for the PR, but it should go to https://github.com/OCA/l10n-vietnam

PR #108, opened 2021-09-09
[14.0][MIG] currency_rate_inverted

AdriaGForgeFlow:
Standard migration of module currency_rate_inverted to v14.

cc ~ @ForgeFlow

PR #107, opened 2021-08-08
[13.0][FIX]Multicompany behaviour fix

SkiBY:
As we have company_dependent attribute, when we don't force context change, there will can be an error if you use multicompany.

All the convertions if user is in another active company will be done by 'current' company and don't like it should be via attributes.

PR #109, opened 2021-10-06
[14.0][MIG] currency_rate_inverted: Migration to 14.0

tupaq:


currency_rate_inverted/__manifest__.py
```diff
@@ -6,7 +6,7 @@
 
 {
     "name": "Currency Rate Inverted",
-    "version": "13.0.1.0.1",
+    "version": "14.0.1.0.0",
```
alexey-pelykh, line 9: This version change should be done as a separate commit. See https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-14.0

alexey-pelykh (changes requested):

currency_rate_inverted/__manifest__.py
tupaq: I already corrected

SkiBY:
As behavior is the same in v.14, you can get changes from https://github.com/OCA/currency/pull/107 directly to this PR

MiquelRForgeFlow:
This should be closed, as the module was already merged in https://github.com/OCA/currency/pull/108.

PR #114, opened 2021-10-21
[12.0][IMP] currency_rate_update_oxr: optional end-of-day exchange rates

pankk:
Adds an option to the settings to use the end-of-day exchange rates for openexchangerates.org.
This is to prevent mismatching currencies when using multi-company and for any reason pulling of exchange rates is delayed. Such a delay would cause today's latest value to be used, which can then be different across different provider records (in the case of multi-company). The (optional) alternative proposed here will use yesterday's end-of-day rates for today's rates.

During testing I had an issue where my local version of Odoo would create a company with currency euro instead of us dollar. I've added a slight change as well to avoid tests breaking when that happens.

### DIFF
--- a/currency_rate_update_oxr/models/res_company.py
+++ b/currency_rate_update_oxr/models/res_company.py
@@ -10,3 +10,7 @@ class ResCompany(models.Model):
     openexchangerates_app_id = fields.Char(
         string='OpenExchangeRates.org App ID',
     )
+    openexchangerates_eod_rates = fields.Boolean(
+        string='End-of-day rates',
+        help='Yesterday\'s end-of-day rates will be used for today.',
+    )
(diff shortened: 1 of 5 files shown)


currency_rate_update_oxr/models/res_company.py
```diff
@@ -10,3 +10,7 @@ class ResCompany(models.Model):
     openexchangerates_app_id = fields.Char(
         string='OpenExchangeRates.org App ID',
     )
+    openexchangerates_eod_rates = fields.Boolean(
```
alexey-pelykh, line 13: I would suggest adding this option to parent module as `use_last_end_of_day_rates` or similar, since in theory not only OXR could have such option (to reduce number of fields)

currency_rate_update_oxr/models/res_currency_rate_provider_OXR.py
```diff
@@ -38,14 +38,15 @@ def _get_supported_currencies(self):
...
         if self.service != 'OXR':
             return super()._obtain_rates(base_currency, currencies, date_from,
                                          date_to)  # pragma: no cover
 
         content = defaultdict(dict)
 
-        date = date_from
-        while date <= date_to:
+        date = date_from - timedelta(days=1) if eod_rates else date_from
+        while date <= (date_to - timedelta(days=1) if eod_rates else date_to):
```
alexey-pelykh, line 49: just alter date_to?

PR #121, opened 2021-11-23
[13.0][FIX] currency_rate_update: Set the correct default value in the company_id field (avoid deprecated options).

victoralmau:
Set the correct default value in the `company_id` field (avoid deprecated options).

This change will need to be applied to v14 as well.

Please @pedrobaeza and @joao-p-marques can you review it?

@Tecnativa TT25624


pedrobaeza (approved):

pedrobaeza:
Fast-tracking the obvious patch:

/ocabot merge patch

See if needed in 14.0

PR #122, opened 2021-11-26
[14.0][FIX] currency_rate_update: Set the correct default value in the company_id field (avoid deprecated options)

victoralmau:
FWP from 13.0: https://github.com/OCA/currency/pull/121

Set the correct default value in the `company_id` field (avoid deprecated options).

This change will need to be applied to v14 as well.

Please @pedrobaeza and @joao-p-marques can you review it?

@Tecnativa TT25624


pedrobaeza (approved):

pedrobaeza:
Another one to fast-track:

/ocabot merge patch

PR #126, opened 2022-01-31
[14.0][ADD] currency_rate_update_nbp

florian-dacosta:
Synchronize rate from the National Bank of Poland.
It is actually kind of a migration from v9/v10.

Still WIP for now

### DIFF
--- a/currency_rate_update_nbp/__init__.py
+++ b/currency_rate_update_nbp/__init__.py
@@ -1,0 +1,1 @@
+from . import models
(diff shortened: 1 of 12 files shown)


alexey-pelykh:
It should go to https://github.com/OCA/l10n-poland

florian-dacosta:
@alexey-pelykh Are you sure about this one ?
All currency_rate_update_* are here as far as I know and it it not directly linked to Poland accounting.

In my case the customer wants to use this rate provider but does not use l10n-poland, it is a french customer, that does a lot of business in Poland.
I would not have thought to check l10n-poland for this but maybe I am wrong.
That beeing said, I have no strong opinion about this

alexey-pelykh:
@florian-dacosta yep, I'm quite sure about this one. It does not need to be directly connected to Polish accounting to live in l10n-poland repository, likewise it contains bank statement import specific for one of Poland banks.

florian-dacosta:
Ok, let's move it then

PR #127, opened 2022-02-23
[14.0][FIX] currency_rate_update: remove obsolete migration script

MiquelRForgeFlow:
They should have been removed when doing the migration.


pedrobaeza (approved):
Indeed

alexey-pelykh (approved):

Issue #136, opened 2022-09-23
Warning on old rates


alexis-via:
In the old days, I used to maintain a module called "currency_rate_date_check" (https://github.com/OCA/account-financial-tools/tree/8.0/currency_rate_date_check) that would raise an error when doing a currency conversion with a rate that was too far away from the date requested by the currency conversion. I stopped maintaining this module for some good reasons... but it was long ago and I don't remember exactly the reasons !!! One of the problems of this module was that, as it was raising an error, it was blocking the process executed by the user, so solving the problem was immediately a very high emergency, which is annoying for the support team.

In the last month, on Odoo v14, I faced failures of currency_rate_update on customer installations: the cron failed to update the rates, but nobody was informed about the problem. From my experience, sending notifications by email for such problems is not a solution ; the mail may be considered as spam, it may arrive in the mailbox of an employee that left the company, etc.
As you know, I like warning banners (cf my overdue warning modules https://github.com/OCA/credit-control/tree/14.0/account_invoice_overdue_warn). So my idea is to have a warning banner on the accounting dashboard when the last currency rate is older than 4 days (4 days will be an  ir.config_parameter).

What do you think of this idea ? Do you have a better idea ?
If you think the idea is good, I'll implement the feature as an OCA module.
If you think it's not so good, then I'll develop it as an Akretion community module or add it to my module account_usability (https://github.com/akretion/odoo-usability/tree/14.0/account_usability).

alexis-via:
My idea to display a warning banner on the accounting dashboard was not good: warning banners are for form views, not for kanban views! 
So I'm thinking about the best way to send a message for users that they can't miss. Create an activity ? Send a chat message ?

nibraun:
I think this is a very good idea ;-) The chat message or simply an email...

alexis-via:
Other ideas include:
- using web_notify https://github.com/OCA/web/tree/14.0/web_notify
- using web_environment_ribbon https://github.com/OCA/web/tree/14.0/web_environment_ribbon
- using custom JS code inspired from web_environment_ribbon or web_switch_context_warning (but I would like to avoid having a module with JS code to maintain)

alexis-via:
web_notify with sticky=True seems like a good option for me. The README of the module for v14 says "To work properly, the server must be launched in gevent mode.", but, according to my tests on v14, it works even when the server is NOT in multi-worker mode (for me, gevent requires multi-worker, but I may be wrong).

rousseldenis:
[...]

PR #137, opened 2022-09-23
[14.0][ADD] module currency_old_rate_notify

alexis-via:
This module uses web_notify to send a notification to accounting managers when currency rates are older than 3 days (3 days is configurable via an ir.config_parameter).

See https://github.com/OCA/currency/issues/136 for the discussion about the design of the module.

notify_old_rates


desdelinux:
@alexis-via I think you should add a parameter to exclude currencies

currency_old_rate_notify/models/res_currency_rate.py
```diff
@@ -0,0 +1,130 @@
...
+        return max_days
+
+    @api.model
+    def notify_rates_too_old(self):
+        max_days = self._notify_rates_get_max_days()
+        today = fields.Date.context_today(self)
+        limit_date = today - timedelta(max_days)
+        logger.debug("notify_rates_too_old limit_date=%s", limit_date)
+        companies = self.env["res.company"].search([])
+        currencies = self.env["res.currency"].search([])
```
paulius-sladkevicius: As it was stated, I agree that needs to filter only active currencies.

paulius-sladkevicius (changes requested):

paulius-sladkevicius:
Great addon - looking forward to trying to use it in production.

nibraun (approved):
Successfully used in prod

desdelinux (approved):

currency_old_rate_notify/models/res_currency_rate.py
alexis-via: that's exactly what the code does :)

paulius-sladkevicius (approved):

andrius-preimantas:
Great module. Any chances having it merged to master? Would like to migrate to v13

klodr:
Tested and approved 
please merge

PR #147, opened 2022-12-12
[13.0][IMP] currency_rate_update: Do not call the _schedule_next_run() method if there is no data.

victoralmau:

@Tecnativa TT40688

### DIFF
--- a/currency_rate_update/models/res_currency_rate_provider.py
+++ b/currency_rate_update/models/res_currency_rate_provider.py
@@ -112,9 +112,6 @@ class ResCurrencyRateProvider(models.Model):
                 [("name", "in", provider._get_supported_currencies())]
             )
 
-    def _get_close_time(self):
-        return False
-
     def _update(self, date_from, date_to, newest_only=False):
         Currency = self.env["res.currency"]
         CurrencyRate = self.env["res.currency.rate"]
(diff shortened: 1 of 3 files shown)


currency_rate_update/models/res_currency_rate_provider.py
```diff
@@ -112,9 +112,6 @@ def _compute_available_currency_ids(self):
                 [("name", "in", provider._get_supported_currencies())]
             )
 
-    def _get_close_time(self):
```
pedrobaeza, line 115: Why removing this?

pedrobaeza: I mean, on weekends, there's no data, so are you sure this is going to work?

victoralmau: Is the approach correct now?

pedrobaeza: If you don't explain more what you want to achieve, I'm not sure.

victoralmau:
Example of use case:

1- We run cron on Friday 10:00 and the API returns no records.
- The `last_successful_run` field is NOT updated.
- The `next_run` field IS updated.

2- We run the cron on Friday 18:00 and the API does return logs:
- The `last_successful_run` field IS updated.
- The `next_run` field IS updated.

Let's assume that Saturday is today's date.
3- We run the cron on Saturday at 10:00 and the API does not return logs:
- The `last_successful_run` field is NOT updated.
- The `next_run` field IS updated.

4- We run the cron on Saturday at 10:00 and the API returns no logs:
- The `last_successful_run` field is NOT updated.
- The `next_run` field IS updated (not being greater than today's date).

This avoids that if the cron is executed and there are **no records**, the data of that day will **never be obtained again**.

pedrobaeza:
But to be completed, you have to iterate through each of the days between `last_successful_run` and `next_run`.

@alexis-via @alexey-pelykh what do you think about this solution?

victoralmau:
> But to be completed, you have to iterate through each of the days between `last_successful_run` and `next_run`.
> 
The current behavior is already like this.

pedrobaeza (approved):

alexey-pelykh (approved):

victoralmau:
Can you try bot process again?

PR #148, opened 2022-12-20
[14.0][IMP] currency_rate_update: Do not call the _schedule_next_run() method if there is no data.

victoralmau:
FWP from 13.0: https://github.com/OCA/currency/pull/147

@Tecnativa TT40688


pedrobaeza (approved):
Still Travis? We should switch to GH actions.

alexey-pelykh (approved):

PR #150, opened 2022-12-20
[15.0][IMP] currency_rate_update: Do not call the _schedule_next_run() method if there is no data.

victoralmau:
FWP from 14.0: https://github.com/OCA/currency/pull/148

@Tecnativa TT40688


pedrobaeza (approved):
Pre-commit should be fixed

victoralmau:
> Pre-commit should be fixed

The pre-commit error is not related (I tried to update the copier template but it doesn't resolve either). Any suggestions?

victoralmau:
Pre-commit error has been solved.

alexey-pelykh (approved):

PR #152, opened 2022-12-21
[16.0][FIX] currency_rate_update

AungKoKoLin1997:
Use context_today() instead of today() to search provider records against next_run.

Before this commit, scheduled updates before 09:00 a.m. would have trouble identifying the target providers for installations in Japan, for example.
**Note**: Japan timezone is GMT+9.


yostashiro (approved):

yostashiro:
@victoralmau @pedrobaeza Could you please review this?

pedrobaeza (approved):
/ocabot merge patch

PR #153, opened 2022-12-22
[16.0][IMP] currency_rate_update: Do not call the _schedule_next_run() method if there is no data.

victoralmau:
FWP from 15.0: https://github.com/OCA/currency/pull/150

@Tecnativa TT40688

### DIFF
--- a/currency_rate_update/models/res_currency_rate_provider.py
+++ b/currency_rate_update/models/res_currency_rate_provider.py
@@ -115,9 +115,6 @@ class ResCurrencyRateProvider(models.Model):
                 [("name", "in", provider._get_supported_currencies())]
             )
 
-    def _get_close_time(self):
-        return False
-
     def _update(self, date_from, date_to, newest_only=False):
         Currency = self.env["res.currency"]
         CurrencyRate = self.env["res.currency.rate"]
@@ -159,8 +156,7 @@ class ResCurrencyRateProvider(models.Model):
                 continue
 
             if not data:
-                if is_scheduled:
-                    provider._schedule_next_run()
+                # Try again if there is no data yet
                 continue
             if newest_only:
                 data = [max(data, key=lambda x: fields.Date.from_string(x[0]))]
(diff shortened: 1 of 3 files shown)


pedrobaeza (approved):

currency_rate_update/models/res_currency_rate_provider.py
```diff
@@ -201,14 +198,20 @@ def _update(self, date_from, date_to, newest_only=False):
...
-        self.last_successful_run = self.next_run
-        self.next_run = (
-            datetime.combine(self.next_run, time.min) + self._get_next_run_period()
-        ).date()
+        self.next_run = min(
+            (
+                datetime.combine(self.next_run, time.min) + self._get_next_run_period()
+            ).date(),
+            fields.Date.today(),
+        )
```
yostashiro: Sorry, I am not getting the rationale of this change. Doesn't this practically revoke the scheduled update interval settings by always setting today's date?

victoralmau: I think this explanation may help to understand it: https://github.com/OCA/currency/pull/147#issuecomment-1346322636

yostashiro: @victoralmau I read it and I think this logic fails on the pattern 2 - `next_run` does not get updated with this logic if I'm not mistaken.

currency_rate_update/models/res_currency_rate_provider.py
```diff
@@ -160,7 +157,8 @@ def _update(self, date_from, date_to, newest_only=False):
 
             if not data:
                 if is_scheduled:
-                    provider._schedule_next_run()
+                    # Try again if there is no data yet
+                    provider.next_run = fields.Date.today()
```
pedrobaeza: Isn't the next_run already in the past?

victoralmau: Use case:
- Next run: 2019-01-01
- We run cron and NO data is returned. Next run: today
(the idea is to retry again according to the frequency set in cron)

Another possibility would be NOT to update next run, what do you think?

pedrobaeza: We shouldn't IMO. Use case:

- Next run: 2022-12-01, Thursday
- We run cron on 2022-12-02 and the webservice is not available.
- Putting next run = 2022-12-02 will lose data for 2022-12-01

Or am I wrong?

victoralmau: Ok, you are right, it is already changed.

yostashiro (approved):
Thanks for the update. Tested locally. 👍

PR #159, opened 2022-12-26
[16.0][IMP] currency_rate_update

kanda999:
Current behavior:
https://github.com/OCA/currency/pull/153 
With this update, If there are days with no data then the rate updates will get stuck.  Get the data, otherwise scheduled_date will not be updated to avoid lost data.
Therefore, you cannot use a provider that does not update data on weekends.

This comment shows that the weekend update with no data was considered in V13.
https://github.com/OCA/currency/pull/147#issuecomment-1346322636

But when there is no data, the method provider._schedule_next_run() has been erased in V16 and backport this update.
https://github.com/OCA/currency/pull/153#discussion_r1055358952

I created a Integer field in provider to be able to use providers with no data on weekends with this update.
Integer field allows you to set the number of tries scheduled action.

### DIFF
--- a/currency_rate_update/models/res_currency_rate_provider.py
+++ b/currency_rate_update/models/res_currency_rate_provider.py
@@ -65,6 +65,7 @@ class ResCurrencyRateProvider(models.Model):
     next_run = fields.Date(
         string="Next scheduled update", default=fields.Date.today, required=True
     )
+    daily = fields.Boolean(compute="_compute_daily", store=True)
 
     _sql_constraints = [
         (
@@ -115,6 +116,13 @@ class ResCurrencyRateProvider(models.Model):
                 [("name", "in", provider._get_supported_currencies())]
             )
 
+    @api.depends("interval_type", "interval_number")
+    def _compute_daily(self):
+        for provider in self:
+            provider.daily = False
+            if provider.interval_type == "days" and provider.interval_number == 1:
+                provider.daily = True
+
     def _update(self, date_from, date_to, newest_only=False):
         Currency = self.env["res.currency"]
         CurrencyRate = self.env["res.currency.rate"]
@@ -161,8 +169,11 @@ class ResCurrencyRateProvider(models.Model):
             if newest_only:
                 data = [max(data, key=lambda x: fields.Date.from_string(x[0]))]
 
+            newest_date = False
             for content_date, rates in data:
                 timestamp = fields.Date.from_string(content_date)
+                if not newest_date or timestamp > newest_date:
+                    newest_date = timestamp
                 for currency_name, rate in rates.items():
                     if currency_name == provider.company_id.currency_id.name:
                         continue
@@ -197,16 +208,17 @@ class ResCurrencyRateProvider(models.Model):
                         )
 
             if is_scheduled:
-                provider._schedule_last_successful_run()
-                provider._schedule_next_run()
+                provider._schedule_last_successful_run(newest_date)
+                provider._schedule_next_run(newest_date)
 
-    def _schedule_last_successful_run(self):
-        self.last_successful_run = self.next_run
+    def _schedule_last_successful_run(self, newest_date):
+        self.last_successful_run = newest_date
 
-    def _schedule_next_run(self):
+    def _schedule_next_run(self, newest_date):
+        # next_run is not used when daily is true, but we are updating the value anyway.
         self.ensure_one()
         self.next_run = (
-            datetime.combine(self.next_run, time.min) + self._get_next_run_period()
+            datetime.combine(newest_date, time.min) + self._get_next_run_period()
         ).date()
 
     def _process_rate(self, currency, rate):
@@ -260,11 +272,14 @@ class ResCurrencyRateProvider(models.Model):
         _logger.info("Scheduled currency rates update...")
 
         today = fields.Date.context_today(self)
+        # When daily is true, the provider should always be picked for scheduled update.
         providers = self.search(
             [
                 ("company_id.currency_rates_autoupdate", "=", True),
                 ("active", "=", True),
+                "|",
                 ("next_run", "<=", today),
+                ("daily", "=", True),
             ]
         )
         if providers:
@@ -278,9 +293,13 @@ class ResCurrencyRateProvider(models.Model):
                     if provider.last_successful_run
                     else (provider.next_run - provider._get_next_run_period())
                 )
+                newest_only = True
                 date_to = provider.next_run
-                provider._update(date_from, date_to, newest_only=True)
-
+                # Fetch next_run to today data
+                if provider.daily:
+                    newest_only = False
+                    date_to = today
+                provider._update(date_from, date_to, newest_only=newest_only)
         _logger.info("Scheduled currency rates update complete.")
 
     def _get_supported_currencies(self):
--- a/currency_rate_update/tests/test_currency_rate_update.py
+++ b/currency_rate_update/tests/test_currency_rate_update.py
@@ -6,6 +6,7 @@ from datetime import date
 from unittest import mock
 
 from dateutil.relativedelta import relativedelta
+from freezegun import freeze_time
 
 from odoo import fields
 from odoo.tests import tagged
@@ -111,13 +112,13 @@ class TestCurrencyRateUpdate(AccountTestInvoicingCommon):
 
     def test_update_ECB_sequence(self):
         self.ecb_provider.interval_type = "days"
-        self.ecb_provider.interval_number = 1
+        self.ecb_provider.interval_number = 2
         self.ecb_provider.last_successful_run = None
         self.ecb_provider.next_run = date(2019, 4, 1)
 
         self.ecb_provider._scheduled_update()
         self.assertEqual(self.ecb_provider.last_successful_run, date(2019, 4, 1))
-        self.assertEqual(self.ecb_provider.next_run, date(2019, 4, 2))
+        self.assertEqual(self.ecb_provider.next_run, date(2019, 4, 3))
         rates = self.CurrencyRate.search(
             [
                 ("company_id", "=", self.company.id),
@@ -127,8 +128,8 @@ class TestCurrencyRateUpdate(AccountTestInvoicingCommon):
         self.assertEqual(len(rates), 1)
 
         self.ecb_provider._scheduled_update()
-        self.assertEqual(self.ecb_provider.last_successful_run, date(2019, 4, 2))
-        self.assertEqual(self.ecb_provider.next_run, date(2019, 4, 3))
+        self.assertEqual(self.ecb_provider.last_successful_run, date(2019, 4, 3))
+        self.assertEqual(self.ecb_provider.next_run, date(2019, 4, 5))
         rates = self.CurrencyRate.search(
             [
                 ("company_id", "=", self.company.id),
@@ -141,7 +142,7 @@ class TestCurrencyRateUpdate(AccountTestInvoicingCommon):
 
     def test_update_ECB_weekend(self):
         self.ecb_provider.interval_type = "days"
-        self.ecb_provider.interval_number = 1
+        self.ecb_provider.interval_number = 2
         self.ecb_provider.last_successful_run = None
         self.ecb_provider.next_run = date(2019, 7, 1)
 
@@ -152,13 +153,25 @@ class TestCurrencyRateUpdate(AccountTestInvoicingCommon):
         self.ecb_provider._scheduled_update()
 
         self.assertEqual(self.ecb_provider.last_successful_run, date(2019, 7, 5))
-        self.assertEqual(self.ecb_provider.next_run, date(2019, 7, 6))
+        self.assertEqual(self.ecb_provider.next_run, date(2019, 7, 7))
 
         self.ecb_provider._scheduled_update()
         self.ecb_provider._scheduled_update()
 
         self.assertEqual(self.ecb_provider.last_successful_run, date(2019, 7, 5))
-        self.assertEqual(self.ecb_provider.next_run, date(2019, 7, 6))
+        self.assertEqual(self.ecb_provider.next_run, date(2019, 7, 7))
+
+    @freeze_time("2022-04-19 22:00")
+    def test_update_ECB_with_daily(self):
+        self.ecb_provider.interval_type = "days"
+        self.ecb_provider.interval_number = 1
+        # Setting last_successful_run a date more than 90 days prior to the freeze_time date,
+        # to make sure that the rates are taken from '/eurofxref-hist.xml'.
+        self.ecb_provider.last_successful_run = date(2021, 4, 19)
+
+        self.ecb_provider._scheduled_update()
+
+        self.assertEqual(self.ecb_provider.last_successful_run, date(2022, 4, 19))
 
     def test_foreign_base_currency(self):
         self.company.currency_id = self.chf_currency
--- a/currency_rate_update/views/res_currency_rate_provider.xml
+++ b/currency_rate_update/views/res_currency_rate_provider.xml
@@ -29,7 +29,8 @@
                 <field name="update_schedule" />
                 <field name="available_currency_ids" invisible="1" />
                 <field name="currency_ids" widget="many2many_tags" />
-                <field name="next_run" />
+                <field name="daily" invisible="1" />
+                <field name="next_run" attrs="{'invisible': [('daily', '=', True)]}" />
             </tree>
         </field>
     </record>
@@ -79,7 +80,11 @@
                             <field name="last_successful_run" />
                         </group>
                         <group>
-                            <field name="next_run" />
+                            <field name="daily" invisible="1" />
+                            <field
+                                name="next_run"
+                                attrs="{'invisible': [('daily', '=', True)]}"
+                            />
                         </group>
                         <group>
                             <field name="available_currency_ids" invisible="1" />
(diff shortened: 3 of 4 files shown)


currency_rate_update/models/res_currency_rate_provider.py
```diff
@@ -65,6 +65,7 @@ class ResCurrencyRateProvider(models.Model):
     next_run = fields.Date(
         string="Next scheduled update", default=fields.Date.today, required=True
     )
+    no_data_skip = fields.Boolean(default=False)
```
yostashiro:
```suggestion
    force_update_next_run = fields.Boolean(
        "Force Update Next Scheduled Update",
        help="Update 'Next scheduled update' to today even when where is no rate data received from the provider in the update process. Select this if the provider does not provide the rate data for certain days (e.g. public holidays)."
    )
```

currency_rate_update/models/res_currency_rate_provider.py
```diff
@@ -156,8 +157,10 @@ def _update(self, date_from, date_to, newest_only=False):
                 continue
 
             if not data:
-                # Try again if there is no data yet
+                if provider.no_data_skip:
+                    provider._schedule_next_run()
```
yostashiro:
```suggestion
                if provider.force_update_next_run:
                    provider.next_run = fields.Date.context_today(self)
```

We should not go as far as running _schedule_next_run() if there is no data.  We should just update next_run to today.

currency_rate_update/models/res_currency_rate_provider.py
```diff
@@ -156,8 +157,10 @@ def _update(self, date_from, date_to, newest_only=False):
...
                 continue
+
```
yostashiro:
```suggestion
```

currency_rate_update/views/res_currency_rate_provider.xml
```diff
@@ -30,6 +30,7 @@
                 <field name="available_currency_ids" invisible="1" />
                 <field name="currency_ids" widget="many2many_tags" />
                 <field name="next_run" />
+                <field name="no_data_skip" />
```
yostashiro:
```suggestion
                <field name="force_update_next_run" optional="hide" />
```

currency_rate_update/views/res_currency_rate_provider.xml
```diff
@@ -89,6 +90,7 @@
                                 options="{'no_create_edit': True}"
                                 domain="[('id', 'in', available_currency_ids)]"
                             />
+                            <field name="no_data_skip" />
```
yostashiro:
```suggestion
                            <field name="force_update_next_run" />
```

yostashiro (changes requested):
Please add tests as well.

yostashiro:
@kanda999 We may also change the tag from 'FIX' to 'IMP' as the issue may not necessarily be a bug.

pedrobaeza:
Why do you need this? European Bank provider, which is the one we use, doesn't update rates on weekends, so current behavior has been though with it in mind.

currency_rate_update/models/res_currency_rate_provider.py
```diff
@@ -65,7 +65,8 @@ class ResCurrencyRateProvider(models.Model):
     next_run = fields.Date(
         string="Next scheduled update", default=fields.Date.today, required=True
     )
-
+    allow_no_data_occurrence = fields.Integer()
+    no_data_occurrence = fields.Integer(readonly=True)
```
yostashiro:
```suggestion
    retry = fields.Integer("Current Try", readonly=True, help="Number of update attempts done by scheduler for the Next Scheduled Update date.")
    max_retry = fields.Integer("Max Retries", help="Maximum number of update attempts for the Next Scheduled Update date. An attempt that goes over this number will update the Next Scheduled Update date to today.")
```

currency_rate_update/models/res_currency_rate_provider.py
```diff
@@ -156,8 +157,13 @@ def _update(self, date_from, date_to, newest_only=False):
                 continue
 
             if not data:
-                # Try again if there is no data yet
+                # Try again if there is no data yet up to set number of times.
+                provider.no_data_occurrence += 1
+                if provider.no_data_occurrence > provider.allow_no_data_occurrence:
+                    provider.next_run = fields.Date.context_today(self)
+                    provider.no_data_occurrence = 0
```
yostashiro:
```suggestion
                if is_sheduled and provider.retry > provider.max_retry:
                    provider.next_run = fields.Date.context_today(self)
                    provider.retry = 0
```

Add the following at the beginning of the loop:

```
if is_scheduled:
    provider.retry += 1
```

currency_rate_update/models/res_currency_rate_provider.py
```diff
@@ -156,8 +157,13 @@ def _update(self, date_from, date_to, newest_only=False):
...
                 continue
+            provider.no_data_occurrence = 0
```
yostashiro:
```suggestion
```

Add `provider.retry = 0` to the end of `_schedule_next_run()`.

kanda999:
@pedrobaeza 
European Bank provider will be stuck at the weekend with current behavior.
You can see that from the test code.
https://github.com/OCA/currency/blob/fbc243cb8a85ffcba5f50d1893fc6c55cdc4bc28/currency_rate_update/tests/test_currency_rate_update.py#L154-L161
In this test
The `next_run` set to Next run : 2019-07-06 is still Next run 2019-07-06 when we run cron. This means it is stuck on the weekend.

So I create integer fields and user can set number of retries in case of no data.

currency_rate_update/views/res_currency_rate_provider.xml
```diff
@@ -89,6 +90,7 @@
                                 options="{'no_create_edit': True}"
                                 domain="[('id', 'in', available_currency_ids)]"
                             />
+                            <field name="allow_no_data_occurrence" />
```
yostashiro:
```suggestion
                            <div>
                                <label for="retry" string="Current Try / Max Retries"/>
                                <field name="retry" class="oe_inline"/> /
                                <field name="max_retry" class="oe_inline"/>
                            </div>
```

yostashiro (changes requested):
In an attempt to make the concept a bit easier to understand...

pedrobaeza:
> The next_run set to Next run : 2019-07-06 is still Next run 2019-07-06 when we run cron. This means it is stuck on the weekend.

But that's the idea. On Monday, it will try to fetch Saturday, Sunday and Monday, and being able to fetch Monday's data, then `next_run` will be changed to Monday.

Why is this? Because with the previous code, if the cron in programmed on Monday, but before Monday data is available, you will lose such rate and not being able to get it again.

So I think you shouldn't reintroduce the old behavior.

kanda999:
@pedrobaeza 
Please see:
https://github.com/OCA/currency/blob/fbc243cb8a85ffcba5f50d1893fc6c55cdc4bc28/currency_rate_update/models/res_currency_rate_provider.py#L281-L282

In the current code, the range to fetch data is up to the date of `next_run`. (`date_to = provider.next_run`)　So if `next_run` is set to Saturday, and the cron runs on Monday, it will not be able to fetch Monday's data and this update stacks.

So we need to improve how many retries we can do like my PR, or change the range of data it can fetch.

pedrobaeza:
I was asking the same in https://github.com/OCA/currency/pull/147#issuecomment-1352700663 and @victoralmau (he's on holidays now) told me that it iterates.

kanda999:
> I was asking the same in [#147 (comment)](https://github.com/OCA/currency/pull/147#issuecomment-1352700663) and @victoralmau (he's on holidays now) told me that it iterates.

Yes, I see that comment.  But now it can not iterate.  I think it's because the logic has changed in https://github.com/OCA/currency/pull/153

pedrobaeza:
But the patch should be then to iterate from next_run to today, not to restore the old behavior.

yostashiro:
@pedrobaeza In that case, do you suggest that `newest_only=True` should be removed from the following line?
https://github.com/OCA/currency/blob/fbc243cb8a85ffcba5f50d1893fc6c55cdc4bc28/currency_rate_update/models/res_currency_rate_provider.py#L282

If you somehow fail to fetch the rate of Monday and today is Tuesday, then you will not get the rate of Monday (you will only get the rate of Tuesday), otherwise.

Edit: if we remove `newest_only=True`, then it is an issue for non-daily intervals, I suppose.

pedrobaeza:
But why adding a new parameter. The system should fetch the remaining dates from `next_run` to today, and set `next_run` on the last day that fetches data. If that's not the behavior, then let's fix it, but not to complicate it with further parameters.

yostashiro:
> But why adding a new parameter.

Because it got lost in translation in this update https://github.com/OCA/currency/pull/153. :sweat_smile:

> The system should fetch the remaining dates from `next_run` to today, and set `next_run` on the last day that fetches data. If that's not the behavior, then let's fix it, but not to complicate it with further parameters.

OK, I'd suggest that we change the following part
https://github.com/OCA/currency/blob/fbc243cb8a85ffcba5f50d1893fc6c55cdc4bc28/currency_rate_update/models/res_currency_rate_provider.py#L281-L282
to
```python
                newest_only = True
                date_to = provider.next_run
                if provider.interval_type == "days" and provider.interval_type == 1:
                    newest_only = False
                    date_to = fields.Date.context_today(self)
                provider._update(date_from, date_to, newest_only=newest_only)
```

This way, I think the behavior should become like how you describe (except for the "and set `next_run` on the last day that fetches data" part, which I think you meant `last_successful_run` perhaps).  Let us know if this makes sense.

@kanda999 Can you go ahead and check if above suggestion works.

pedrobaeza:
Well, the goal is to not lose a day rate if something goes bad on the daily cron. The previous case with ECB was that they publish the rates at 17:00, so if you run the cron before that date, internal data was changed and no more possibility to get the rate for that date.

If the algorithm should take `last_successful_run` or `next_run`, I don't know right now. I wonder also why 2 variables are needed at the end. I would say that fetching from `last_successful_run` to today, and setting `last_successful_run` on the last fetched date would be enough.

kanda999:
> If the algorithm should take `last_successful_run` or `next_run`, I don't know right now. I wonder also why 2 variables are needed at the end. I would say that fetching from `last_successful_run` to today, and setting `last_successful_run` on the last fetched date would be enough.

Users do not necessarily want to get data every day..  User can set interval_type and interval_number.  next_run have the date until the next time get data, so next_run is needed for the variable.

I updated @yostashiro suggestion.  It follows when interval_type is days and interval_number is 1.

pedrobaeza (approved):
Just a thought: do you really think someone is fetching rates by other interval than day?

currency_rate_update/models/res_currency_rate_provider.py
```diff
@@ -198,15 +201,15 @@ def _update(self, date_from, date_to, newest_only=False):
 
             if is_scheduled:
                 provider._schedule_last_successful_run()
```
yostashiro:
```suggestion
                provider._schedule_last_successful_run(newest_date)
```

currency_rate_update/models/res_currency_rate_provider.py
```diff
@@ -198,15 +201,15 @@ def _update(self, date_from, date_to, newest_only=False):
...
-                provider._schedule_next_run()
+                provider._schedule_next_run(newest_date)
 
     def _schedule_last_successful_run(self):
         self.last_successful_run = self.next_run
```
yostashiro:
```suggestion
    def _schedule_last_successful_run(self, newest_date):
        self.last_successful_run = newest_date
```

currency_rate_update/tests/test_currency_rate_update.py
```diff
@@ -152,13 +152,24 @@ def test_update_ECB_weekend(self):
...
+    def test_update_ECB_by_every_days(self):
+        self.ecb_provider.interval_type = "days"
+        self.ecb_provider.interval_number = 1
+        self.ecb_provider.last_successful_run = None
+        self.ecb_provider.next_run = date(2019, 7, 5)
+
+        self.ecb_provider._scheduled_update()
+
+        self.assertEqual(self.ecb_provider.last_successful_run, date(2019, 7, 5))
+        self.assertEqual(self.ecb_provider.next_run, self.today)
```
yostashiro: I suppose we cannot rely on this test and we should remove this.  In fact, we should not use/update `next_run` at all if the provider is configured for daily updates.

yostashiro (changes requested):

yostashiro:
> Just a thought: do you really think someone is fetching rates by other interval than day?

We are in no position to be able to judge this.  We have internal practice of setting rates monthly in our company but are not using this module.

currency_rate_update/models/res_currency_rate_provider.py
```diff
@@ -197,16 +201,17 @@ def _update(self, date_from, date_to, newest_only=False):
...
+                provider._schedule_next_run(newest_date)
 
-    def _schedule_last_successful_run(self):
-        self.last_successful_run = self.next_run
+    def _schedule_last_successful_run(self, newest_date):
+        self.last_successful_run = newest_date
 
-    def _schedule_next_run(self):
+    def _schedule_next_run(self, newest_date):
+        # If you set daily update, fetch to today date.
```
yostashiro:
```suggestion
        # next_run is not used when daily is true, but we are updating the value anyway.
```

currency_rate_update/models/res_currency_rate_provider.py
```diff
@@ -260,11 +272,14 @@ def _scheduled_update(self):
         _logger.info("Scheduled currency rates update...")
 
         today = fields.Date.context_today(self)
+        # When you set daily, you should get data up to today.
```
yostashiro:
```suggestion
        # When daily is true, the provider should always be picked for scheduled update.
```

currency_rate_update/models/res_currency_rate_provider.py
```diff
@@ -65,6 +65,7 @@ class ResCurrencyRateProvider(models.Model):
     next_run = fields.Date(
         string="Next scheduled update", default=fields.Date.today, required=True
     )
+    daily = fields.Boolean(compute="_compute_daily_provider", store=True)
```
yostashiro:
```suggestion
    daily = fields.Boolean(compute="_compute_daily", store=True)
```

currency_rate_update/models/res_currency_rate_provider.py
```diff
@@ -115,6 +116,13 @@ def _compute_available_currency_ids(self):
                 [("name", "in", provider._get_supported_currencies())]
             )
 
+    @api.depends("interval_type", "interval_number")
+    def _compute_daily_provider(self):
```
yostashiro:
```suggestion
    def _compute_daily(self):
```

currency_rate_update/tests/test_currency_rate_update.py
```diff
@@ -152,13 +153,26 @@ def test_update_ECB_weekend(self):
...
         self.assertEqual(self.ecb_provider.last_successful_run, date(2019, 7, 5))
-        self.assertEqual(self.ecb_provider.next_run, date(2019, 7, 6))
+        self.assertEqual(self.ecb_provider.next_run, date(2019, 7, 7))
+
+    @freeze_time("2022-04-19 22:00")
+    def test_update_ECB_with_daily(self):
+        self.ecb_provider.interval_type = "days"
+        self.ecb_provider.interval_number = 1
+        # If date renge is not more than 90 days, you get an error.
+        # Because defferent URL is used in ECB.
```
yostashiro:
```suggestion
        # Setting last_successful_run a date more than 90 days prior to the freeze_time date,
        # to make sure that the rates are taken from '/eurofxref-hist.xml'.
```

currency_rate_update/tests/test_currency_rate_update.py
```diff
@@ -152,13 +153,26 @@ def test_update_ECB_weekend(self):
...
+        self.ecb_provider.last_successful_run = date(2021, 4, 19)
+        self.ecb_provider.next_run = date(2022, 4, 19)
```
yostashiro:
```suggestion
```

currency_rate_update/views/res_currency_rate_provider.xml
```diff
@@ -29,7 +29,8 @@
                 <field name="update_schedule" />
                 <field name="available_currency_ids" invisible="1" />
                 <field name="currency_ids" widget="many2many_tags" />
-                <field name="next_run" />
+                <field name="next_run" attrs="{'invisible': [('daily', '=', True)]}"/>
+                <field name="daily" invisible="1" />
```
yostashiro:
```suggestion
                <field name="daily" invisible="1" />
                <field name="next_run" attrs="{'invisible': [('daily', '=', True)]}"/>
```

currency_rate_update/views/res_currency_rate_provider.xml
```diff
@@ -79,7 +80,8 @@
                             <field name="last_successful_run" />
                         </group>
                         <group>
-                            <field name="next_run" />
+                            <field name="next_run" attrs="{'invisible': [('daily', '=', True)]}" />
+                            <field name="daily" invisible="1" />
```
yostashiro:
```suggestion
                            <field name="daily" invisible="1" />
                            <field name="next_run" attrs="{'invisible': [('daily', '=', True)]}" />
```

yostashiro (changes requested):

yostashiro:
@kanda999 Please squash commits and add proper commit message.

yostashiro (approved):
LGTM. 👍

yostashiro:
@pedrobaeza Can you please review this again as the code has been updated since your approval?

pedrobaeza:
OK, in a quick glance seems correct. Tell us when you apply it to your instances to backport it.

PR #161, opened 2023-01-02
[add] Croatia HNB rates

badbole:
As of 1.1.2023 Croatia is using EURO as official currency.
This module is a getter for HNB mid rate .


alexey-pelykh:
Country-specific modules should go to country-specific repositories, [l10n-croatia](https://github.com/OCA/l10n-croatia) in this case

badbole:
> Country-specific modules should go to country-specific repositories, [l10n-croatia](https://github.com/OCA/l10n-croatia) in this case

I was not sure where to make this PR... 
currency repository seems more appropriate, as this is not localisation related.
i see other currency rate provider modules in lower version branches... 
so... i still belive this repo is the right one for this PR... 

can we have other opinions? i will gladly move this PR to l10n-croatia, 

@alexis-via @feketemihai and others.. can we have other thoughts on this matter?

Leave PR open for currency repo or move it ot localisation repo?

pedrobaeza:
I would go for the localization one being country-specific. Anyway, isn't now you under the ECB?

badbole:
@pedrobaeza tnx for answer... i will take your opinion as qualified and move this PR to l10-croatia... 
for info... we stared this year with EUR as new currency. but we still need to apply Croatia national bank (HNB) as currency rate provider. I'm still not sure will there be diffecence in rates between HNB and ECB, but currently, 
HNB rates are to be used.   so... suggestions accepted, closing this PR and moving it to l10n-coratia.

pedrobaeza:
Great. The main reason to move it to a localization repo is to not depend on PSC that doesn't know about the specifics of your country. It also allows you to discuss in your native language instead of English.

PR #162, opened 2023-01-03
[13.0][FIX] currency_rate_update

victoralmau:
Backport from 16.0: https://github.com/OCA/currency/commit/ecd222d2e0eadc1bc0ae17dde3005cbbccb225d3

Before this commit, there was a problem of next_run getting stuck when there is no rate data received for the day (e.g. weekend).

This commit fixes the issue by totally ignoring next_run for daily update settings, meaning that the provider always fetches the latast rates from last_successful_run to today.

Please @pedrobaeza can you review it?

@Tecnativa TT40688

PR #163, opened 2023-01-03
[14.0][FIX] currency_rate_update

victoralmau:
FWP from 13.0: https://github.com/OCA/currency/pull/162

Before this commit, there was a problem of next_run getting stuck when there is no rate data received for the day (e.g. weekend).

This commit fixes the issue by totally ignoring next_run for daily update settings, meaning that the provider always fetches the latast rates from last_successful_run to today.

@Tecnativa TT40688


test-requirements.txt
```diff
@@ -0,0 +1 @@
+freezegun
```
yostashiro: We can remove this file.  freezegun is already part of Odoo core in 14.0.

yostashiro (approved):
👍

pedrobaeza (approved):
/ocabot merge minor

PR #164, opened 2023-01-05
[15.0][FIX] currency_rate_update

victoralmau:
FWP from 14.0: https://github.com/OCA/currency/pull/163

Before this commit, there was a problem of next_run getting stuck when there is no rate data received for the day (e.g. weekend).

This commit fixes the issue by totally ignoring next_run for daily update settings, meaning that the provider always fetches the latast rates from last_successful_run to today.

@Tecnativa TT40688

### DIFF
--- a/currency_rate_update/models/res_currency_rate_provider.py
+++ b/currency_rate_update/models/res_currency_rate_provider.py
@@ -65,6 +65,7 @@ class ResCurrencyRateProvider(models.Model):
     next_run = fields.Date(
         string="Next scheduled update", default=fields.Date.today, required=True
     )
+    daily = fields.Boolean(compute="_compute_daily", store=True)
 
     _sql_constraints = [
         (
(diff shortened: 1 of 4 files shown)


pedrobaeza (approved):

currency_rate_update/models/res_currency_rate_provider.py
```diff
@@ -278,9 +293,22 @@ def _scheduled_update(self):
...
-                provider._update(date_from, date_to, newest_only=True)
-
+                current_utc_hour = datetime.now().hour
+                _logger.debug(
+                    "Provider %s date_to=%s today=%s, current hour %s UTC",
+                    provider.name,
+                    date_to,
+                    today,
+                    current_utc_hour,
+                )
```
yostashiro, line 305: I don't see these lines in other versions, but I am guessing that it is an intended change for logging purpose.

yostashiro (approved):

currency_rate_update/models/res_currency_rate_provider.py
victoralmau: This code currently exists in v14 (https://github.com/OCA/currency/blob/14.0/currency_rate_update/models/res_currency_rate_provider.py#L300-L307)
It was added at https://github.com/OCA/currency/pull/135 to 14.0 but was not ported to v15.

PR #173, opened 2023-06-09
[16.0] currency_rate_update: fix bad strings

alexis-via:
I found these bad string while translating the module to French.
Also, no need to set global=True on ir.rule, because it is a computed field, https://github.com/odoo/odoo/blob/16.0/odoo/addons/base/models/ir_rule.py#L52

### DIFF
--- a/currency_rate_update/models/res_currency_rate_provider.py
+++ b/currency_rate_update/models/res_currency_rate_provider.py
@@ -32,7 +32,7 @@ class ResCurrencyRateProvider(models.Model):
     )
     active = fields.Boolean(default=True)
     service = fields.Selection(
-        string="Source Service",
+        string="Provider",
         selection=[("none", "None")],
         default="none",
         required=True,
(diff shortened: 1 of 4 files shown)


rvalyi (approved):

currency_rate_update/models/res_currency_rate_provider.py
```diff
@@ -52,31 +52,31 @@ class ResCurrencyRateProvider(models.Model):
...
+        string="Next Scheduled Update", default=fields.Date.today, required=True
     )
     daily = fields.Boolean(compute="_compute_daily", store=True)
 
     _sql_constraints = [
         (
             "service_company_id_uniq",
             "UNIQUE(service, company_id)",
-            "Service can only be used in one provider per company!",
+            "This provider has already been setup in this company.",
```
yostashiro, line 74:
```suggestion
            "This provider has already been set up in this company.",
```

yostashiro (approved):

PR #216, opened 2026-05-13
[16.0][FIX] currency_rate_update_xe: use XE.com JSON API (HTML scraping returns 403)

DarioLodeiros:
### Problem

XE.com now answers automated requests to its public HTML currency tables
(`www.xe.com/currencytables`) with a CloudFront **403 "Request blocked"**,
regardless of the `User-Agent` sent. This breaks `_obtain_rates`, which
scrapes that page: the scheduled update fails every run and posts a
"Currency Rate Provider Failure" message, leaving rates frozen.

The earlier mitigation in this PR (sending a `User-Agent` and raising on
HTTP errors) is no longer enough — the block is applied at the edge for
datacenter IP ranges, so no header tweak gets through.

### Fix

Switch the provider to the JSON endpoint that powers xe.com's own
converter (`/api/protected/midmarket-converter/`), which is reachable
from datacenter hosts. It returns the latest USD-based mid-market rates;
these are converted to be relative to the company base currency.

Since the endpoint only exposes the latest rates, the provider now
always returns today's rates regardless of the requested range, which is
sufficient for the scheduled daily update.

### Notes

- `_get_supported_currencies` is unchanged (the API uses the same ISO codes).
- Verified end-to-end against the live endpoint from a datacenter host
  (DOP-based company, USD rate): scraping returns 403, the JSON API
  returns 200 with correct pegs (e.g. AED 3.6725, BHD 0.376).

### DIFF
--- a/currency_rate_update_xe/__manifest__.py
+++ b/currency_rate_update_xe/__manifest__.py
@@ -3,7 +3,7 @@
 
 {
     "name": "Currency Rate Update: XE.com",
-    "version": "16.0.1.0.0",
+    "version": "16.0.1.1.0",
     "category": "Financial Management/Configuration",
     "summary": "Update exchange rates using XE.com",
     "author": "Tecnativa, Odoo Community Association (OCA)",
--- a/currency_rate_update_xe/models/res_currency_rate_provider_XE.py
+++ b/currency_rate_update_xe/models/res_currency_rate_provider_XE.py
@@ -1,13 +1,18 @@
 # Copyright 2023 Tecnativa - Ernesto Tejeda
 # License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html).
 
-from datetime import date, timedelta
+import base64
+from datetime import date
 
 import requests
-from lxml import etree
 
 from odoo import _, fields, models
 from odoo.exceptions import UserError
+
+# Public credentials embedded in xe.com's own front-end, required to reach its
+# JSON converter endpoint.
+XE_API_URL = "https://www.xe.com/api/protected/midmarket-converter/"
+XE_API_TOKEN = base64.b64encode(b"lodestar:pugsnax").decode()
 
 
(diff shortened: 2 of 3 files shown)


DarioLodeiros:
Pushed a follow-up commit that fixes the underlying cause of the test failure: `_get_historical_rate` was building `?date=YYYY-MM-DD` URLs for every day in the range, but XE.com returns **HTTP 404** when the requested date is today. The scheduled cron path always lands on today (`date_to == today`), and `test_cron` reproduces this exact shape (`date_from = yesterday, date_to = today`).

Before this PR the 404 was silently parsed as empty HTML, so today's rate was not produced but past-day rates from the same iteration were. Now with `raise_for_status()`, the 404 interrupts the whole loop and the past-day rates were also lost — that's what the failing assertion in `test_cron` was catching.

The fix falls back to the latest endpoint (no `date` query string) whenever the iterated day is today or in the future, reusing what `_get_latest_rate` already does.

Note: the `pre-commit` failure is unrelated — `setuptools-odoo-make-default`/`setuptools-odoo-get-requirements` blow up with `ModuleNotFoundError: No module named 'pkg_resources'`, which affects any PR currently opened against this repo (Python 3.13 / setuptools >= 81 removed `pkg_resources`).

DarioLodeiros:
Side note on the red `pre-commit` check: it's caused by `setuptools-odoo` 3.1.8 importing the removed `pkg_resources` module and is unrelated to this PR's diff. Opened #217 to bump the hook to 3.3.2 (which other OCA 16.0 repos already use). Once #217 lands and this branch is rebased, pre-commit should go green.

bashlk:
Tried several other fixes to the currency_rate_update_xe module from other PRs but this is the only one that worked. Thanks @DarioLodeiros.

Can also confirm the XE.com blocks requests to `/currencytables` from my cloud provider (Hetzner) but the endpoint used in this PR still works.

PR #219, opened 2026-06-19
[14.0][IMP] currency_rate_update_wise: rename from transferwise + security hardening

klodr:
## currency_rate_update_wise (rename of currency_rate_update_transferwise)

TransferWise rebranded to **Wise** in 2021. This renames the module `currency_rate_update_transferwise` → `currency_rate_update_wise` and the provider `TransferWise` → `Wise`, with identical rate-fetching functionality.

- New provider `res_currency_rate_provider_Wise.py`.
- `migrations/14.0.2.0.0/pre-migration.py` migrates existing `res.currency.rate.provider` records (service `TransferWise` → `Wise`) and the related `ir.model.data`, so existing configurations keep working after a plain module update.
- Security hardening of the rate-fetch HTTP call: request timeout, query-param encoding, explicit handling of non-JSON responses, and the API key restricted to `base.group_system`.
- Hermetic tests (HTTP mocked), i18n (es/fr/it) and README updated.

See the commit message for the full rationale (Wise vs ECB coverage, UUID→JWT token transition).

Migrations to 15.0–18.0 will follow in dedicated PRs.


klodr:
@OCA/currency-maintainers when you have a moment, could you please review? This renames `currency_rate_update_transferwise` → `currency_rate_update_wise` (TransferWise → Wise rebrand) with an OpenUpgrade pre-migration and some security hardening. CI is green except the unrelated `xe` test and the `pre-commit`/`pkg_resources` infra issue, both addressed in #220 and #221. Thanks!

klodr:
### Note for reviewers: the red CI here is not caused by this PR

Both failures come from the 14.0 branch itself, and both already have a fix
open in this repository. Merging them first turns this PR green with no
change to its content.

**`pre-commit` — `ModuleNotFoundError: No module named 'pkg_resources'`**
Raised by the `setuptools-odoo-make-default` and
`setuptools-odoo-get-requirements` hooks: setuptools ≥ 81 dropped
`pkg_resources`. Fixed by #221, which bumps setuptools-odoo to 3.3.2.

**`test with OCB` / `test with Odoo` — `AssertionError: 0 != 1`**
Raised by `currency_rate_update_xe`, in `test_cron` and `test_wizard`. Those
tests query x-rates.com over the network and now get nothing back, so the
whole branch is red regardless of the module under review. Fixed by #220,
which sends a User-Agent, raises on HTTP errors and mocks the tests.

Suggested order: #221, then #220, then this one. Happy to rebase once they
land.

PR #218, opened 2026-05-19
[17.0][FIX] currency_rate_update_xe: send User-Agent and raise on HTTP error

shirashi3771:
Based on #216 .

@mpascuall , @lbarry-apsl , @peluko00, @javierobcn, @BernatObrador , @cubells , @isugac, @palomagrc93, @LlucSG, @ppyczko , @miquelalzanillas please review.

PR #220, opened 2026-06-20
[14.0][FIX] currency_rate_update_xe: send User-Agent, raise on HTTP error, mock tests

klodr:
Backport to 14.0 of the xe.com fix from #216 (16.0) and #218 (17.0), made hermetic.

## Problem
`test_cron` / `test_wizard` fail with `AssertionError: 0 != 1`: xe.com now returns **403** for requests without a `User-Agent`, and **404** on the dated endpoint (`?date=<today>`). The scheduled run always ends on today, so no rate is produced.

## Fix (as discussed in #216 / #218)
- `_request_data`: send a `User-Agent`, add a request timeout and `raise_for_status()` → an HTTP error surfaces as a clean `UserError` instead of being parsed as empty HTML.
- `_get_historical_rate`: fall back to the latest endpoint (no `?date=`) for today/future days, as `_get_latest_rate` already does.

## Tests made hermetic
The tests no longer hit xe.com: `_request_data` is mocked, and the latest/historical fallback, the HTML parsing and the HTTP-error path are all covered. This removes the dependency on the live website — the actual cause of the recurring red CI raised in the comments of #216 / #218.

## Context
This also unblocks the CI of #219 (`currency_rate_update_wise`), currently held red only by these unrelated `xe` failures. `wise` relies on a stable authenticated JSON API rather than HTML scraping, so its own tests are already deterministic.

Refs #216, #218.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -140,7 +140,7 @@ repos:
           - --settings=.
         exclude: /__init__\.py$
   - repo: https://github.com/acsone/setuptools-odoo
-    rev: 3.1.8
+    rev: 3.3.2
     hooks:
       - id: setuptools-odoo-make-default
       - id: setuptools-odoo-get-requirements
--- a/currency_rate_update_xe/__manifest__.py
+++ b/currency_rate_update_xe/__manifest__.py
@@ -3,7 +3,7 @@
 
 {
     "name": "Currency Rate Update: XE.com",
-    "version": "14.0.1.0.0",
+    "version": "14.0.1.0.1",
     "category": "Financial Management/Configuration",
     "summary": "Update exchange rates using XE.com",
     "author": "Tecnativa, Odoo Community Association (OCA)",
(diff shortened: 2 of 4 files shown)


klodr:
@OCA/currency-maintainers @DarioLodeiros could you please review? This backports to 14.0 the xe.com fix from #216/#218 (User-Agent + latest-endpoint fallback) and additionally makes the tests hermetic (HTTP mocked), so they no longer depend on the live website. `test with OCB`/`test with Odoo` are green; the remaining `pre-commit` red is the `pkg_resources` infra issue fixed by #221. Thanks!

klodr:
Folded #221 into this PR, because the two were deadlocking each other:

| | `pre-commit` | `test with Odoo` / `test with OCB` |
|---|---|---|
| this PR (xe fix) | ❌ `setuptools-odoo` → `pkg_resources` ImportError | ✅ ✅ |
| #221 (`setuptools-odoo` 3.1.8 → 3.3.2) | ✅ | ❌ ❌ (the very xe tests this PR fixes) |

Each one fixed the check the other was failing, so neither could ever go green on its own — and since the CI runs the whole repo's test suite, the two failing `currency_rate_update_xe` tests were keeping **every open 14.0 PR** red.

The one-line `.pre-commit-config.yaml` bump from #221 is now the second commit here, so this branch should be green end to end. #221 will be closed as redundant.

@OCA/currency-maintainers — this is the unblocking change for the 14.0 branch; once it lands, the other pending 14.0 PRs should go green on rebase.

PR #221, opened 2026-06-20
[14.0] pre-commit: bump setuptools-odoo to 3.3.2 to fix pkg_resources ImportError

klodr:
`setuptools-odoo` 3.1.8 imports the removed `pkg_resources` module, so the `setuptools-odoo-make-default` and `setuptools-odoo-get-requirements` pre-commit hooks fail with `ModuleNotFoundError: No module named 'pkg_resources'` on recent setuptools (>= 81). The `pre-commit` check is therefore red on every PR currently opened against `14.0`.

Bump the hook to `3.3.2`, the same fix as #217 (16.0).

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -140,7 +140,7 @@ repos:
           - --settings=.
         exclude: /__init__\.py$
   - repo: https://github.com/acsone/setuptools-odoo
-    rev: 3.1.8
+    rev: 3.3.2
     hooks:
       - id: setuptools-odoo-make-default
       - id: setuptools-odoo-get-requirements


klodr:
@OCA/currency-maintainers could you please review? This is the 14.0 counterpart of #217: bump `setuptools-odoo` to 3.3.2 so the pre-commit hooks stop failing with `ModuleNotFoundError: No module named 'pkg_resources'`. `pre-commit` is green here; the test jobs stay red only because of the unrelated `xe` issue (fixed by #220). Thanks!

klodr:
Closing as redundant: the one-line `setuptools-odoo` 3.1.8 → 3.3.2 bump is now the second commit of #220.

The two PRs were deadlocking each other — this one had `pre-commit` green but the `currency_rate_update_xe` tests red, #220 had the opposite. Neither could go green alone, which is why both sat untouched for ~3 months. Combined, #220 is now green on all 7 checks.

Credit for the bump stays on the original commit, which was cherry-picked as-is.

PR #224, opened 2026-06-20
[17.0][MIG] currency_rate_update_wise

klodr:
## `currency_rate_update_wise` → 17.0

Forward-port of the **Wise.com** exchange-rate provider from 14.0 to 17.0.
Framework migration only, no functional change.

### Notes for reviewers
- **Hermetic tests**: the Wise HTTP layer (`_wise_provider_retrieve`) is mocked —
  no network access needed in CI.
- Carries a small, non-blocking helper that logs a warning when the configured
  API token still uses the legacy UUID format (Wise UUID → JWT migration).


klodr:
A note on the red CI, for reviewers:

The only failing checks are **test with OCB / test with Odoo**, and they fail **solely on the pre-existing `currency_rate_update_xe` module** — its `test_cron` / `test_wizard` make real network calls to xe.com (which no longer returns usable data), so they assert `0 != 1`. This is unrelated to this PR.

`currency_rate_update_wise` itself installs cleanly and **all 10 of its tests pass** — the Wise HTTP layer is mocked, so no network access is needed in CI. The xe.com test issue is already addressed for 17.0 in #218 (and #216 for 16.0, #220 for 14.0).

